PR #2, opened 2024-10-16
Optimize Docker Build Layers and Add Sudo Privileges for Fast-LLM Container

tscholak:
I'd like to refine the `Dockerfile` slightly to improve build efficiency and add runtime flexibility for the Fast-LLM container. The changes are small but impactful, focusing on two main improvements:

1. Improved Build Layering for Faster Rebuilds:
    * The build process is now split into two distinct stages:
        1. Dependency installation (based on `setup.py`, `setup.cfg`, `pyproject.toml`) is done first.
        2. Fast-LLM code installation is done last, by using the new `--exclude=` option enabled by Dockerfile syntax version `1.7-labs`.
    * With this change the dependencies don't need to be reinstalled when the Fast-LLM source code changes. That can reduce rebuild times significantly since code changes land in different Docker image layers than dependencies.
2. Added Sudo Privileges for Fast-LLM User:
    * Introduced password-less sudo privileges to the `fast_llm` user. This addition allows system adjustments (e.g., modifying system limits or adjusting host settings) directly from within the container.
    * I found this very useful in bare Kubernetes environments (like LambdaLabs), where I needed to frequently make changes to system configurations (such as those controllable with `ulimit`) that do not persist across container restarts.

Here's a breakdown of the build time:

```
docker build --platform linux/amd64 -t torstenscholak663/fast-llm:latest --build-arg FAST_LLM_USER_ID=1000 .                                   
[+] Building 48.3s (23/23) FINISHED                                                                                                                                                                                                                               docker:desktop-linux
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                              0.0s
 => => transferring dockerfile: 1.48kB                                                                                                                                                                                                                                            0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1.7-labs                                                                                                                                                                                                  0.7s
 => [auth] docker/dockerfile:pull token for registry-1.docker.io                                                                                                                                                                                                                  0.0s
 => CACHED docker-image://docker.io/docker/dockerfile:1.7-labs@sha256:b99fecfe00268a8b556fad7d9c37ee25d716ae08a5d7320e6d51c4dd83246894                                                                                                                                            0.0s
 => [internal] load metadata for nvcr.io/nvidia/pytorch:24.07-py3                                                                                                                                                                                                                 1.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                                 0.0s
 => => transferring context: 163B                                                                                                                                                                                                                                                 0.0s
 => [ 1/15] FROM nvcr.io/nvidia/pytorch:24.07-py3@sha256:f47441c102a810a27758b0b6274d46012ac15fd467119b2e1f0467be82bc8af3                                                                                                                                                         0.0s
 => [internal] load build context                                                                                                                                                                                                                                                 0.0s
 => => transferring context: 12.73kB                                                                                                                                                                                                                                              0.0s
 => CACHED [ 2/15] RUN apt-get update     && apt-get install --no-install-recommends -y git-lfs sudo util-linux     && rm -rf /var/lib/apt/lists/*     && git lfs install                                                                                                         0.0s
 => CACHED [ 3/15] RUN useradd -m -u 1000 -s /bin/bash fast_llm     && echo 'fast_llm ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers                                                                                                                                                    0.0s
 => CACHED [ 4/15] WORKDIR /app                                                                                                                                                                                                                                                   0.0s
 => [ 5/15] COPY --chown=fast_llm ./fast_llm/csrc/ fast_llm/csrc/                                                                                                                                                                                                                 0.1s
 => [ 6/15] RUN make -C ./fast_llm/csrc/                                                                                                                                                                                                                                          4.9s
 => [ 7/15] COPY --chown=fast_llm setup.py setup.cfg ./                                                                                                                                                                                                                           0.0s
 => [ 8/15] RUN PIP_NO_INPUT=1 pip3 install --no-cache-dir ".[CORE,OPTIONAL,DEV]"                                                                                                                                                                                                35.7s
 => [ 9/15] COPY --chown=fast_llm ./Megatron-LM Megatron-LM                                                                                                                                                                                                                       0.0s 
 => [10/15] COPY --chown=fast_llm ./examples examples                                                                                                                                                                                                                             0.0s 
 => [11/15] COPY --chown=fast_llm ./tests tests                                                                                                                                                                                                                                   0.0s 
 => [12/15] COPY --chown=fast_llm ./tools tools                                                                                                                                                                                                                                   0.0s 
 => [13/15] COPY --exclude=./fast_llm/csrc/ --chown=fast_llm ./fast_llm/ fast_llm/                                                                                                                                                                                                0.0s 
 => [14/15] COPY --chown=fast_llm pyproject.toml ./                                                                                                                                                                                                                               0.0s 
 => [15/15] RUN PIP_NO_INPUT=1 pip3 install --no-deps -e .                                                                                                                                                                                                                        4.6s
 => exporting to image                                                                                                                                                                                                                                                            1.0s
 => => exporting layers                                                                                                                                                                                                                                                           1.0s
 => => writing image sha256:f9b20cc3ca3c99ad8d3788cb6eacf5f48d518f48aec3bc3250f8d1d0d7cedeb3                                                                                                                                                                                      0.0s 
 => => naming to docker.io/torstenscholak663/fast-llm:latest                                                                                                                                                                                                                      0.0s 
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,30 +1,38 @@
 FROM nvcr.io/nvidia/pytorch:24.07-py3
 
-# git-lfs is needed to interact with the huggingface hub
+# Install git-lfs for Huggingface hub interaction and sudo for system adjustments
 RUN apt-get update \
-    && apt-get install git-lfs \
+    && apt-get install --no-install-recommends -y git-lfs sudo \
     && rm -rf /var/lib/apt/lists/* \
     && git lfs install
 
+# Add a user for Fast-LLM with sudo privileges for runtime adjustments
 ARG FAST_LLM_USER_ID=1000
+RUN useradd -m -u $FAST_LLM_USER_ID -s /bin/bash fast_llm \
+    && echo 'fast_llm ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
 
-RUN useradd -m -u $FAST_LLM_USER_ID -s /bin/bash -d /home/fast_llm fast_llm
 USER fast_llm
 WORKDIR /app
-ENV PYTHONPATH=/app:/app:/app/Megatron-LM
-ENV PATH=$PATH:/app:/home/fast_llm/.local/bin/
 
-COPY --chown=fast_llm setup.py setup.cfg ./
-COPY --chown=fast_llm fast_llm/__init__.py ./fast_llm/
+# Environment settings for Python and PATH
+ENV PYTHONPATH=/app:/app:/app/Megatron-LM \
+    PATH=$PATH:/home/fast_llm/.local/bin/
 
-RUN PIP_NO_INPUT=1 pip3 install --no-cache-dir -e ".[CORE,OPTIONAL,DEV]"
+# Copy the dependency files and install dependencies
+COPY --chown=fast_llm setup.py setup.cfg pyproject.toml ./
+RUN PIP_NO_INPUT=1 pip3 install --no-cache-dir ".[CORE,OPTIONAL,DEV]"
 
-COPY --chown=fast_llm fast_llm/csrc/ ./fast_llm/csrc/
-RUN make -C ./fast_llm/csrc/
-
+# Copy the rest of the code
 COPY --chown=fast_llm ./Megatron-LM Megatron-LM
 COPY --chown=fast_llm ./examples examples
 COPY --chown=fast_llm ./tests tests
 COPY --chown=fast_llm ./tools tools
-COPY --chown=fast_llm ./fast_llm fast_llm
-COPY --chown=fast_llm fast_llm/tools/train.py pyproject.toml ./
+
+# Compile the C++ extensions (fast_llm/csrc)
+COPY --chown=fast_llm fast_llm/csrc/ ./fast_llm/csrc/
+RUN make -C ./fast_llm/csrc/
+
+# Copy the main source code for Fast-LLM and install in editable mode
+COPY --chown=fast_llm ./fast_llm/ ./fast_llm/ \
+     --exclude ./fast_llm/csrc/
+RUN PIP_NO_INPUT=1 pip3 install --no-deps -e .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1.7-labs
 FROM nvcr.io/nvidia/pytorch:24.07-py3
 
 # Install git-lfs for Huggingface hub interaction and sudo for system adjustments
@@ -29,10 +30,9 @@ COPY --chown=fast_llm ./tests tests
 COPY --chown=fast_llm ./tools tools
 
 # Compile the C++ extensions (fast_llm/csrc)
-COPY --chown=fast_llm fast_llm/csrc/ ./fast_llm/csrc/
+COPY --chown=fast_llm ./fast_llm/csrc/ fast_llm/csrc/
 RUN make -C ./fast_llm/csrc/
 
 # Copy the main source code for Fast-LLM and install in editable mode
-COPY --chown=fast_llm ./fast_llm/ ./fast_llm/ \
-     --exclude ./fast_llm/csrc/
+COPY --exclude=./fast_llm/csrc/ --chown=fast_llm ./fast_llm/ fast_llm/
 RUN PIP_NO_INPUT=1 pip3 install --no-deps -e .
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM nvcr.io/nvidia/pytorch:24.07-py3
 
 # Install git-lfs for Huggingface hub interaction and sudo for system adjustments
 RUN apt-get update \
-    && apt-get install --no-install-recommends -y git-lfs sudo \
+    && apt-get install --no-install-recommends -y git-lfs sudo util-linux \
     && rm -rf /var/lib/apt/lists/* \
     && git lfs install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,13 +19,13 @@ WORKDIR /app
 ENV PYTHONPATH=/app:/app/Megatron-LM \
     PATH=$PATH:/home/fast_llm/.local/bin/
 
-# Compile the C++ extensions (fast_llm/csrc)
-COPY --chown=fast_llm ./fast_llm/csrc/ fast_llm/csrc/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,30 +1,38 @@
...
 COPY --chown=fast_llm ./Megatron-LM Megatron-LM
 COPY --chown=fast_llm ./examples examples
 COPY --chown=fast_llm ./tests tests
 COPY --chown=fast_llm ./tools tools
-COPY --chown=fast_llm ./fast_llm fast_llm
-COPY --chown=fast_llm fast_llm/tools/train.py pyproject.toml ./
+
+# Compile the C++ extensions (fast_llm/csrc)
+COPY --chown=fast_llm ./fast_llm/csrc/ fast_llm/csrc/
+RUN make -C ./fast_llm/csrc/
```
jlamypoirier, line 34: Why moving here? I would much rather keep it where it was because csrc almost never changes and takes way longer than the copies above.

Dockerfile
```diff
@@ -1,30 +1,38 @@
...
+
+# Copy the main source code for Fast-LLM and install in editable mode
+COPY --exclude=./fast_llm/csrc/ --chown=fast_llm ./fast_llm/ fast_llm/
```
jlamypoirier, line 37: Not sure the exclude is worth it here, especially if it means the extra syntax linek.

Dockerfile
```diff
@@ -1,30 +1,38 @@
...
+RUN PIP_NO_INPUT=1 pip3 install --no-deps -e .
```
jlamypoirier, line 38: Why the need for another install here? What was wrong with the previous version?

Dockerfile
```diff
@@ -1,30 +1,38 @@
...
+# Environment settings for Python and PATH
+ENV PYTHONPATH=/app:/app:/app/Megatron-LM \
+    PATH=$PATH:/home/fast_llm/.local/bin/
 
-RUN PIP_NO_INPUT=1 pip3 install --no-cache-dir -e ".[CORE,OPTIONAL,DEV]"
-
-COPY --chown=fast_llm fast_llm/csrc/ ./fast_llm/csrc/
-RUN make -C ./fast_llm/csrc/
+# Copy the dependency files and install dependencies
+COPY --chown=fast_llm setup.py setup.cfg pyproject.toml ./
```
jlamypoirier, line 23: Why adding the toml? It's not used for the installation.

Dockerfile
```diff
@@ -1,30 +1,38 @@
...
+# Install git-lfs for Huggingface hub interaction and sudo for system adjustments
 RUN apt-get update \
-    && apt-get install git-lfs \
+    && apt-get install --no-install-recommends -y git-lfs sudo util-linux \
     && rm -rf /var/lib/apt/lists/* \
     && git lfs install
 
+# Add a user for Fast-LLM with sudo privileges for runtime adjustments
 ARG FAST_LLM_USER_ID=1000
+RUN useradd -m -u $FAST_LLM_USER_ID -s /bin/bash fast_llm \
```
jlamypoirier, line 12: 🥇

jlamypoirier:
>    * With this change the dependencies don't need to be reinstalled when the Fast-LLM source code changes. That can reduce rebuild times significantly since code changes land in different Docker image layers than dependencies.

Not sure I'm following here, was it not the case already?

Dockerfile
tscholak: This is the most important change: In the previous version, dependencies and code were installed at the same time. Now we first install the dependencies (see above) and only make the editable install at the very end. It's also ensured this way that pip can find and link all code in the fast-llm folder.

Dockerfile
tscholak: this is done such that the compiled code isn't overwritten again.

Dockerfile
tscholak: ah, ok, let's move it up then!

Dockerfile
tscholak: it appeared to me that the setup.* files and pyproject.toml fulfil similar purposes and could be grouped together. it is not uncommon to specify dependencies in the pyproject.toml file, because this is how setup tools usually works.

jlamypoirier: We don't do that in fast-llm though, the toml file is just there because black needs it.

tscholak: ok, I'll split it up then and copy pyproject.toml somewhere else

jlamypoirier: I don't really think it's worth copying at all... But if we do keep it let's keep it here so we don't add another line

Dockerfile
tscholak: ok, looks like you are right.

tscholak:
> Not sure I'm following here, was it not the case already?

People were telling me it was not. I never checked those claims, I just reworked the Dockerfile such that it was clear and sure that we wouldn't always rebuild everything on small code changes. Looks like it wasn't truly necessary. I removed those changes.

Dockerfile
```diff
@@ -1,30 +1,37 @@
...
-RUN PIP_NO_INPUT=1 pip3 install --no-cache-dir -e ".[CORE,OPTIONAL,DEV]"
-
-COPY --chown=fast_llm fast_llm/csrc/ ./fast_llm/csrc/
+# Compile the C++ extensions (fast_llm/csrc)
+COPY --chown=fast_llm ./fast_llm/csrc/ fast_llm/csrc/
 RUN make -C ./fast_llm/csrc/
 
+# Copy the dependency files and install dependencies
+COPY --chown=fast_llm setup.py setup.cfg pyproject.toml ./
+RUN PIP_NO_INPUT=1 pip3 install --no-cache-dir -e ".[CORE,OPTIONAL,DEV]"
```
jlamypoirier, line 28: Let's move this back above the crsc compile, since this installation is much longer.

jlamypoirier (commented):
I double-checked the build times, it was ~100-500 ms before this PR even with code changes. I suspect the complains about rebuilding happened because  of the multiple recent changes to dependencies, etc., that do force a re-install.

The rest of this PR looks useful though, I have one minor comment and then we can merge.

Dockerfile
tscholak: done

PR #5, opened 2024-10-17
StarDoc model training

akshaykalkunte:
WIP StarDoc model integration into FastLLM


tscholak:
Hi @jlamypoirier! @akshaykalkunte and I talked and we want to push this PR over the finish line. There's a lot going on here, and we should review the approach top down to decide how this needs to be refactored to go into `main`. At the top of my head are the following separate concerns:

1. Model architecture: Are VLMs GPTs from the point of view of Fast-LLM? I think they aren't because too much is different. We should add a new model architecture (e.g. "vlm") to Fast-LLM.
2. Data preprocessing: Related to #38, we should factor out data preprocessing and introduce an offline preprocessing step, `fast-llm prepare_data vlm --config stardoc.yaml`, that makes `VLMMemmapDataset`s and stores them on disk.
3. Vision encoder implementation: Right now it's a monolithic wrapper layer that uses a HF auto model. We should discuss if and when we reimplement this in Fast-LLM. This can be a separate effort and (as a side effect) result in yet another model class, `vision_encoder`, that we can also train from scratch if we wanted to.
4. Cross-attention instead of adapter layer: StarDoc is moving towards a special form of cross-attention between the vision encoder and the LM decoder. This likely has implications for parallelization.
5. Llama 3 support: StarDoc will use pre-trained Llama 3.2 (text-only?) models, we need to be able to load them. See also #39.

tscholak:
> 4\. Cross-attention instead of adapter layer: StarDoc is moving towards a special form of cross-attention between the vision encoder and the LM decoder. This likely has implications for parallelization.

As @akshaykalkunte pointed out recently, AlignVLM will be the best path forward for this first implementation. I read the paper and I don't see any obstacles. The method is refreshingly simple.

PR #6, opened 2024-10-17
Better checkpoints

jlamypoirier:
* Add a FieldUpdate tool to simplify field overrides.
* Standardize checkpoint saving and loading configs, using mixins so training checkpoint/export configs can use just the right fields (ex. no path since it's provided by the trainer)
* Adjust some fields so they make sense for both saving and loading, and merge the ones for config selection into an enum (see below).
* Move checkpoint save/load from Run to Trainer and simplify.
* Make IntervalConfig into a proper class (use FieldUpdate foe doc instead)
* Make export checkpoint fully configurable, save separately instead of using a symlink.
* Add keep_every for checkpoints to get the old behaviour, and optional callback script.

RENAME[BW COMPATIBLE] pretrained.imported_type ->pretrained.model_type
RENAME[BW COMPATIBLE] pretrained.load_weights -> pretrained.model_weights
RENAME[BW COMPATIBLE] pretrained.load_optimizer -> pretrained.optimizer_state
MERGE (pretrained.override_architecture, pretrained.load_full_base_model_config, pretrained.load_full_fast_llm_config) ->pretrained.load_config:enum

### DIFF
--- a/fast_llm/engine/config_utils/checkpoint.py
+++ b/fast_llm/engine/config_utils/checkpoint.py
@@ -24,7 +24,7 @@ class CheckpointType(str, enum.Enum):
     external = "external"
 
 
-class LoadConfig(str, enum.Enum):
+class ModelConfigType(str, enum.Enum):
     none = "none"
     architecture = "architecture"
     model = "model"
@@ -32,15 +32,15 @@ class LoadConfig(str, enum.Enum):
 
     @property
     def load_architecture(self):
-        return self != LoadConfig.none
+        return self != ModelConfigType.none
 
     @property
     def load_base_model(self):
-        return self in (LoadConfig.model, LoadConfig.fast_llm)
+        return self in (ModelConfigType.model, ModelConfigType.fast_llm)
 
     @property
     def load_fast_llm(self):
-        return self == LoadConfig.fast_llm
+        return self == ModelConfigType.fast_llm
 
 
 @config_class()
@@ -66,20 +66,11 @@ class CheckpointConfigBase(Config):
         desc="Model type for external models (ex. Huggingace model name).",
         hint=FieldHint.feature,
     )
-    load_config: LoadConfig = Field(
-        default=LoadConfig.architecture,
-        desc="Configuration to save/load.",
-        hint=FieldHint.core,
-    )
     fast_llm_config: bool = Field(
         default=False,
         desc="Save/load the full fast-llm model configuration, including the distributed and multi-stage configurations.",
         hint=FieldHint.feature,
     )
-
-    @property
-    def compare_log_fn(self):
-        return ValueError if self.load_config.load_architecture else logger.warning
 
     @classmethod
     def _from_dict(
@@ -131,20 +122,30 @@ class CheckpointSaveConfigBase(Config):
 
 
 @config_class()
-class CheckpointMetadataConfig(CheckpointPathConfigBase, CheckpointConfigBase):
+class CheckpointSaveMetadataConfig(CheckpointPathConfigBase, CheckpointConfigBase):
     _abstract = False
 
 
 @config_class()
-class CheckpointSaveConfig(CheckpointMetadataConfig, CheckpointStateConfigBase, CheckpointSaveConfigBase):
+class CheckpointSaveConfig(CheckpointSaveMetadataConfig, CheckpointStateConfigBase, CheckpointSaveConfigBase):
     _abstract = False
 
 
 @config_class()
-class CheckpointLoadConfig(CheckpointMetadataConfig, CheckpointStateConfigBase):
+class CheckpointLoadMetadataConfig(CheckpointPathConfigBase, CheckpointConfigBase):
     _abstract = False
 
+    load_config: ModelConfigType = Field(
+        default=ModelConfigType.architecture,
+        desc="Configuration to save/load.",
+        hint=FieldHint.core,
+    )
 
-# @config_class()
-# class TrainingExportConfig(CheckpointConfigBase, CheckpointStateConfigBase, CheckpointSaveConfigBase):
-#    _abstract=False
+    @property
+    def compare_log_fn(self):
+        return ValueError if self.load_config.load_architecture else logger.warning
+
+
+@config_class()
+class CheckpointLoadConfig(CheckpointLoadMetadataConfig, CheckpointStateConfigBase):
+    _abstract = False
--- a/fast_llm/engine/config_utils/checkpoint.py
+++ b/fast_llm/engine/config_utils/checkpoint.py
@@ -2,6 +2,7 @@
 import enum
 import logging
 import pathlib
+import typing
 import warnings
 
 from fast_llm.config import Config, Field, FieldHint, check_field, config_class
@@ -15,7 +16,7 @@ CHECKPOINT_VERSION = "0.1"
 KNOWN_CHECKPOINT_VERSIONS = ("0", "0.1")
 
 
-class CheckpointType(str, enum.Enum):
+class CheckpointFormat(str, enum.Enum):
     # Distributed checkpoint for fast checkpointing and resuming.
     distributed = "distributed"
     # Model state dict, for safe long-term storage in Fast-LLM format.
@@ -56,8 +57,8 @@ class CheckpointPathConfigBase(Config):
 @config_class()
 class CheckpointConfigBase(Config):
     _abstract = True
-    format: CheckpointType = Field(
-        default=CheckpointType.distributed,
+    format: CheckpointFormat = Field(
+        default=CheckpointFormat.distributed,
         desc="Format of the checkpoint.",
         hint=FieldHint.core,
     )
@@ -70,14 +71,14 @@ class CheckpointConfigBase(Config):
     @classmethod
     def _from_dict(
         cls,
-        default: dict[str],
+        default: dict[str, typing.Any],
         strict: bool = True,
         flat: bool = False,
     ):
         # TODO v0.2: Remove.
         if default.get("format", None) == "huggingface":
             warnings.warn(f"`huggingface` checkpoint format has been renamed to `external`.")
-            default["format"] = CheckpointType.external.value
+            default["format"] = CheckpointFormat.external.value
         cls._handle_renamed_field(default, "imported_type", "model_type")
         return super()._from_dict(default, strict, flat)
 
@@ -91,7 +92,7 @@ class CheckpointStateConfigBase(Config):
     @classmethod
     def _from_dict(
         cls,
-        default: dict[str],
+        default: dict[str, typing.Any],
         strict: bool = True,
         flat: bool = False,
     ):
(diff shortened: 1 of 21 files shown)


fast_llm/engine/config_utils/checkpoint.py
```diff
@@ -0,0 +1,147 @@
...
+# TODO: Use packaging.version? (Safer but extra requirement)
+CHECKPOINT_VERSION = "0.1"
+KNOWN_CHECKPOINT_VERSIONS = ("0", "0.1")
+
+
+class CheckpointFormat(str, enum.Enum):
+    # Distributed checkpoint for fast checkpointing and resuming.
+    distributed = "distributed"
+    # Model state dict, for safe long-term storage in Fast-LLM format.
+    state_dict = "state_dict"
```
tscholak, line 23: how about we call this `long_term` instead?

fast_llm/engine/config_utils/checkpoint.py
```diff
@@ -0,0 +1,147 @@
...
+    def _from_dict(
+        cls,
+        default: dict[str, typing.Any],
+        strict: bool = True,
+        flat: bool = False,
+    ):
+        # TODO v0.2: Remove.
+        if default.get("format", None) == "huggingface":
+            warnings.warn(f"`huggingface` checkpoint format has been renamed to `external`.")
+            default["format"] = CheckpointFormat.external.value
```
tscholak, line 81: will there be other external checkpoint formats or why this change?

fast_llm/engine/config_utils/checkpoint.py
```diff
@@ -0,0 +1,147 @@
+# TODO: Use packaging.version? (Safer but extra requirement)
```
tscholak, line 1: not sure where to ask this question:
where will we be configuring the layers-per-step option?

fast_llm/engine/config_utils/checkpoint.py
```diff
@@ -0,0 +1,147 @@
...
+logger = logging.getLogger(__name__)
+
+# TODO: Use packaging.version? (Safer but extra requirement)
+CHECKPOINT_VERSION = "0.1"
+KNOWN_CHECKPOINT_VERSIONS = ("0", "0.1")
+
+
+class CheckpointFormat(str, enum.Enum):
+    # Distributed checkpoint for fast checkpointing and resuming.
+    distributed = "distributed"
```
tscholak, line 21: how about we call this `fast` or `short_term`?

fast_llm/engine/multi_stage/fast_llm_model.py
```diff
@@ -83,28 +85,83 @@ def distributed(self):
...
+            "checkpoint_type": CheckpointFormat.distributed.value,
+            "checkpoint_version": str(CHECKPOINT_VERSION),
+            "fast_llm_config": self._fast_llm_config.to_serialized(),
+            "state_shard_names": list(self._state_shard_names[:num_shards]),
+            "metadata": {} if metadata is None else metadata,
+        }
+
+        # TODO: Simplify branching.
+        if checkpoint_config.format == CheckpointFormat.external:
+            # TODO: Support optimizer?
```
tscholak, line 104: can you confirm that Fast-LLM will set the `checkpoint_config.optimizer_state` to `False` somewhere (or reject the config at launch) before we end up here? Would be frustrating to crash thousands of steps into training because of this.

fast_llm/engine/multi_stage/fast_llm_model.py
```diff
@@ -83,28 +85,83 @@ def distributed(self):
...
+            ):
+                shard_split = self._state_shard[i].split(self._stage_shard_sizes, 0)
+                for stage, shard in zip(self._stages_on_device.values(), shard_split):
+                    for name, tensor in stage._export_shard(shard, dtype=checkpoint_config.data_type):  # noqa
+                        assert name not in fast_llm_state_dict
+                        fast_llm_state_dict[(name, shard_name)] = tensor
+                        for exported_name, exported_tensor in converter.convert_state_dict(
+                            fast_llm_state_dict, True
+                        ).items():
+                            context.add_tensor(exported_name, exported_tensor)
```
tscholak, line 155: so "external" is just state-dict saving with ad-hoc conversion?

in that case, do we need "external" as its own format?

right now:

1. `distributed`: fast internal checkpoint format, only for short-term use (i.e. within the same run and with the same (minor?) version of Fast-LLM)
2. `state_dict`: long-term internal checkpoint format, safe to use with different versions of Fast-LLM
3. `external`: like `state_dict`, but potentially lossy because optimizer states cannot be saved and tensors may be converted in a numerically lossy way.

did I capture that correctly?

fast_llm/engine/training/config.py
```diff
@@ -17,48 +25,85 @@
...
-            default=0,
-            desc=f"Offset for the first {offset_desc or desc}.",
-            hint=FieldHint.feature,
-            valid=check_field(Assert.geq, 0),
+@config_class()
+class IntervalConfig(Config):
+    # Intervals are a common pattern, so we standardize them with this base class.
+    interval: int | None = Field(
+        default=None,
+        desc="The number of training iterations between each interval. Setting to None will disable.",
```
tscholak, line 33: maybe the wrong place to do this, but can we make it so that there is at least a warning that checkpoint saving is disabled? I don't think people will appreciate training for hours (or days) only to find out that they forgot to set a saving interval ;)

fast_llm/engine/training/config.py
```diff
@@ -17,48 +25,85 @@
...
+            environment.update(self.environment)
+            subprocess.Popen(self.script, env=environment)
+
+
+@config_class()
+class WandbAlertConfig(IntervalConfig):
+    interval = FieldUpdate(
+        desc="The number of training iterations between each Wandb status post (alert)."
+        " Setting to None will disable iteration-based wandb alerts."
+        " Must be a sub-interval of the logging interval."
```
tscholak, line 102: you mean that wandb posting can only happen at logging times? That would be super-interval, no? can you clarify what you mean here? I'm confused.

fast_llm/engine/training/config.py
```diff
@@ -90,71 +141,109 @@ class WandbConfig(Config):
...
-        )
+        return (self.get_count(training_iterations) + extra_validations) * self.iterations if self.enabled() else 0
 
 
 @config_class()
-class CheckpointConfig(get_interval_config_class("checkpoint")):
+class CheckpointBaseConfig(IntervalConfig):
+    _abstract = True
+    save_name: typing.ClassVar[str] = "save"
+    directory_name: typing.ClassVar[str] = "save"
```
tscholak, line 166: for what kind of checkpoint will we have `save` as the name of the output directory?

fast_llm/engine/training/config.py
```diff
@@ -90,71 +141,109 @@ class WandbConfig(Config):
...
-    callback: CallbackConfig = Field(
-        default_factory=CallbackConfig,
-        desc="Callback (shell script) to run after export.",
-        hint=FieldHint.core,
+class ExportConfig(CheckpointBaseConfig, CheckpointConfigBase, CheckpointStateConfigBase, CheckpointSaveConfigBase):
+    _abstract = False
+    save_name: typing.ClassVar[str] = "export"
+    directory_name = "export"
+    interval = FieldUpdate(
+        desc="The number of training iterations between each export." " Setting to None will disable exports."
```
tscholak, line 228: Can this interval be incompatible with the checkpointing interval (i.e., not an exact multiple or divisor of the checkpointing interval)?

fast_llm/engine/training/config.py
```diff
@@ -90,71 +141,109 @@ class WandbConfig(Config):
...
 
 
 @config_class()
-class ShutdownConfig(get_interval_config_class("automated shutdown")):
-    pass
+class ShutdownConfig(IntervalConfig):
+    interval = FieldUpdate(
+        desc="The number of training iterations between each automated shutdown."
+        " Setting to None will disable automated shutdowns."
+        " Must be a sub-interval of the checkpoint interval."
```
tscholak, line 242: what's an automated shutdown? I am not familiar with this feature? is this new?

tscholak (commented):
Great changes, thanks @jlamypoirier.
I have a few suggestions and questions for clarification. Can you please answer them?

fast_llm/engine/config_utils/checkpoint.py
jlamypoirier: That's a hack for the single process converter, it doesn't apply here because we already have the whole model loaded. I'm not entirely sure about memory usage but it should be OK because we reconstruct one layer at the time.

fast_llm/engine/config_utils/checkpoint.py
jlamypoirier: Also an option, let's discuss it later.

fast_llm/engine/multi_stage/fast_llm_model.py
jlamypoirier: Right now there is no safety check, but that would be easy to add. Looking at it, right now the bigger problem is that the optimizer isn't saved by default so the bigger risk is not saving it when we want to...

fast_llm/engine/multi_stage/fast_llm_model.py
jlamypoirier: Something like that, I'll be clearing things up in the next PR. The plan is to generalize things a bit, something like
```
- AbstractSaver
    - DistributedSaver [hard-coded]
    - AbstractStateDictSaver
        - StateDictSaver [hard-coded]
        - HuggingfaceSaver
            - [model/user-specific]
        - [model/user-specific]
    - [model/user-specific]

jlamypoirier: As for the optimizer state, the only reason it's prevented is I haven't implemented it for huggingface, it would probably not be that hard.

fast_llm/engine/training/config.py
jlamypoirier: Yes. By sub-interval I meant posting iterations are a subset of logging ones, but that's probably not the right term.

fast_llm/engine/training/config.py
jlamypoirier: None, that's a placeholder (I could just remove?)

fast_llm/engine/training/config.py
jlamypoirier: Yes it doesn't matter anymore because checkpoints and exports are completely independent.

fast_llm/engine/training/config.py
jlamypoirier: It's always been there but never used.

fast_llm/engine/training/config.py
jlamypoirier: That's an option, but would the warning actually be useful? A runtime warning is unlikely to be seen, and a validation one is only useful if the config is validated before launch.

tscholak: Hm, I see. Do you have a better idea?

tscholak (approved):
Thanks @jlamypoirier for answering my comments. Since most changes I suggested will be future work, this can be merged as is.

fast_llm/engine/training/config.py
jlamypoirier: Not really....

PR #7, opened 2024-10-17
Fix static call

nitsanluke:


tools/push_model.py
```diff
@@ -72,7 +72,7 @@ def _get_commited_iter_numbers(hf_repo: hf_hub.Repository) -> list[int]:
         print("commits", commits)
         commits = [c.split(" ", 1)[1] for c in commits]
         # Keep commits corresponding to a new iter
-        return [iter_number for c in commits if (iter_number := self.get_iter_number(c)) is not None]
+        return [iter_number for c in commits if (iter_number := PushConfig._get_iter_number(c)) is not None]
```
tscholak, line 75: minor nitpick: this should probably a class method instead of a static method. Then:

```suggestion
        return [iter_number for c in commits if (iter_number := cls._get_iter_number(c)) is not None]
```

tscholak (approved):
LGTM

tools/push_model.py
nitsanluke: @tscholak cls not defined? or are you suggesting to change `_get_commited_iter_numbers` to class method and use self?

nitsanluke: Thanks made the method into a classmethod.

PR #8, opened 2024-10-19
Create CODE_OF_CONDUCT.md

tscholak:
- [x] adapt https://github.com/ServiceNow/azimuth/blob/main/CODE_OF_CONDUCT.md for Fast-LLM
- [x] create CITATION.cff
- [ ] create fast-llm-team@servicenow.com email account

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+### ServiceNow Open Source Code-of-Conduct
+
+This code of conduct provides guidelines for participation in ServiceNow-managed open-source communities and projects.
+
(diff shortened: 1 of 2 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,46 @@
+### ServiceNow Open Source Code-of-Conduct
```
jlamypoirier, line 1: Some of these don't make much sense here, I'm assuming t's servicenow boilerplate?

tscholak: We need to ask Sean what needs to be included. Got all of this from Azimuth

PR #71, opened 2024-11-27
Fix small typos in the help and tutorial

chrish42:
* Quickstart is not walking the user through setting up Docker, so don't say it's doing that
* Mention Kubeflow more
*  Hugging Face -> HuggingFace


tscholak (approved):
LGTM!!!

PR #104, opened 2025-01-06
Modular dataset configuration

jlamypoirier:
# ✨ Description

A set of composable and dynamic dataset configuration classes, that allow defining arbitrary dataset definition schemes.

Dynamic configuration classes: experimental implementation in `GPTDatasetConfig`. if it works well we could use elsewhere, ex. for defining plugins.. It defines a class registry that is populated in `__init_subclass__`, and works as long as the subclass is imported.

The data config now has a `dataset` entry, which can be any dataset config. That dataset config may contain further nested dataset configs, etc. 
Datasets can be sampled or samplable. Both support `build_and_sample(config)`, but samplable (subclass of sampled) also supports `build()`. Indexed datasets are a special case of samplable datasets which also support indexing (get and __len__).

The types for now are:
* `memmap`: A typical Megatron dataset, indexed
* `concatenated`: The concatenation of multiple indexed datasets if it were one. Currently unused.
* `slice`: A contiguous slice of an indexed dataset, for subsampling or train/valid/test split.
* `blended`: Blend sampled datasets according to the given probabilities.
* `dummy`: Always returns the same sample. Only available as sampled.
* `legacy`: Same as before this PR, for backward compatibility only. This is the only way to do dataset from json files, which we aim to replace with a concatenated one anyway.

Datasets are defined are `data.datasets.[Phase]`
Dataset classes may (and typically do) include nested dataset definitions.

Misc:
* Remove split dataset machinery.

Breaking change: `sample` dataset source has been dropped since it's not that relevant. Otherwise configs are backward-compatible (for now).

For future work:
* Concatenate dataset from all files in directory.
* Remove capitalization on phase names in config
* Make phase names more flexible? (plain str instead of enum)
* `shuffle` wrapper for indexed datasets?
* Move dataset definition next to phase definition in train config?
* More features.
* Simplify things?

## 🔍 Type of change

Select all that apply:


tscholak:
Hi @jlamypoirier,  

I took a closer look at the dynamic registration of class members, particularly the use of `type_` as a keyword argument in the class definition. I think I understand the intention behind the design, and it strongly reminds me of what otherwise is known as a tagged union or sum type. I think it's a powerful abstraction for handling modular configurations, and I like it. However, the use of `type_` as a keyword argument directly in the class definition feels very unconventional for Python. I think such a mechanism could be better handled using explicit class-level attributes or fields. Are you familiar with how Pydantic supports tagged unions? I'm not suggesting we adopt Pydantic directly, but I think it provides a useful reference for building a more canonical solution.

I doodled a bit and came up with an alternative approach inspired by Pydantic's tagged unions but generalized using a metaclass. With it, the `DatasetConfig` could look like this:

```python
class DatasetConfig(TaggedUnionBase):
    __tag_field__ = "kind"  # Configurable tag field name

class MemmapConfig(DatasetConfig):
    kind: ClassVar[str] = "memmap"
    path: str

class ConcatenatedConfig(DatasetConfig):
    kind: ClassVar[str] = "concatenated"
    datasets: list[DatasetConfig]

class SliceConfig(DatasetConfig):
    kind: ClassVar[str] = "slice"
    dataset: DatasetConfig
```

With this structure, the machinery for tagged unions is centralized in a `TaggedUnionBase` class, making the implementation cleaner. Instead of using a keyword argument in the class definition (which I think is very non-standard for Python), we use here a `ClassVar` (e.g., `kind`) to declare the type. The `kind` field then acts as a discriminator for serialization and deserialization, and new types can be seamlessly added as subclasses.

An example implementation of the metaclass powering this mechanism can look like this:

```python
class TaggedUnionMeta(type):
    _registries: dict[Type, dict[str, Type]] = {}

    def __new__(cls, name: str, bases: tuple[Type, ...], dct: dict):
        tag_field = dct.pop("__tag_field__", "type")
        cls_obj = super().__new__(cls, name, bases, dct)
        if bases:  # Skip the base class itself
            base = bases[0]
            cls._registries.setdefault(base, {})
            cls._registries[base][getattr(cls_obj, tag_field, name.lower())] = cls_obj
            cls_obj._tag_field = tag_field
        return cls_obj

class TaggedUnionBase(metaclass=TaggedUnionMeta):
    @classmethod
    def from_dict(cls, data: dict):
        tag_value = data.pop(cls._tag_field)
        return TaggedUnionMeta._registries[cls][tag_value](**data)
```

I think this approach nicely encapsulates the tagging mechanism into a reusable base class, which then keeps the configuration code clean. Like this PR's implementation based on `__init_subclass__`, this implementation provides straightforward extensibility: new types register themselves automatically without requiring modifications to existing code.

jlamypoirier:
@tscholak That's another option, it's basically the same thing done differently.

The choice between `__init_subclass__` is mostly personal preference here. Both do the same thing, I think metaclasses are more common just because `__init_subclass__` is relatively recent so wasn't an option in many cases. Do you have a specific reason for using a metaclass?

 I personally prefer `__init_subclass__` because it's way easier to deal with, also a metaclass could interfere with other ones (ex. abc.ABC). 

Concerning the kwarg for the class, it's not necessary, but I chose to have it for clarity and to deal with edge cases. The main idea is we could have a complicated hierarchy of classes, and we don't want to add all of them to the registry (ex. abstract/generic ones). With a plain classvar it's harder to tell whether we want to add the class to the registry, because the value could have been inherited from the base class. There are obviously solution without the class kwargs, but I thought it was the cleanest way.

jlamypoirier:
Hi Torsten! Thanks for your comments, but I must respectfully disagree on some aspects.

The `metaclass` vs `__init_subclass__` and `kwarg` vs `Classvar` are independent questions, let's treat them as such.

`__init_subclass__` is the new(ish) simpler and better way to register subclasses since python 3.6 https://peps.python.org/pep-0487/, I think this alone should settle it as per our contribution guidelines.

I don't really get either of your arguments for a metaclass, in fact I see them as arguments for the opposite. Metaclasses are a complicated and obscure concept most developers should never have to deal with. Yes some experienced developers may have worked with them in the past, but they will have no problem dealing with the simpler version. In my opinion most developers won't know either, and `__init_subclass__` is way more accessible. Therefore, in terms of readability, maintainability, and accessibility, `__init_subclass__` is clearly better.

> With the metaclass approach (or something else that leads to sum-type patterns similar to Pydantic), contributors only need to understand:
> 
> - A base class representing the sum type.
> - Subclasses with a `ClassVar` acting as a type tag.
> 
> This is simpler, more explicit, and easier to discover.

There is absolutely no difference with `__init_subclass__`. The `kwarg` vs `Classvar` thing is a separate concern.

I don't see much value trying to imitate pydantic:
* There is no difference from the user point of view, other than `__init_subclass__` avoiding potential metaclass interactions. One would have to dig deep to notice the difference in implementation.
* It's not really a design choice. Pydantic's development started before python 3.6, so they had to use metaclass. They might also have other reasons for using metaclasses that don't apply here.

jlamypoirier:
As for the separate `kwarg` vs `Classvar` matter, this is really a design choice, we can discuss it discuss it in our next meeting. I introduced the kwarg to try it out, because I think it may offer a good alternative to some of the messier parts of Fast-LLM, including the massive use of class variables. Yes I'm aware of the alternative ways to do it, the `abstract` flag is already there actually (though I'd prefer a better solution).

tscholak:
Hi again,  

I think we may still be talking past each other, so I'd like to clarify my position.  

I really don't care whether we use `__init_subclass__`, a metaclass, or something else for the internal mechanism. That's an implementation detail, and it can live in a part of the codebase nobody ever touches again. The point I'm raising is about the **interface contributors use to define and tag new configuration classes**. This is what I care about:  

1. No contributor should need to understand the internal mechanics to define config classes. Whether it's `__init_subclass__` or not behind the scenes, it should be abstracted away. Nobody should need to dig into this unless they are maintaining that very core mechanism. Pydantic is a good example: nobody looks at its metaclass implementation, yet the interface is clear and consistent.

2. The tagging mechanism must be clear, conventional, and Pythonic. Using kwargs in class definitions isn't something any Python developers will expect. I've never seen this being used before, and I suspect others haven't either. Contributors need to understand how to define new configuration classes and tag them in a way that feels natural. Class-level attributes are well-known and widely used for tagging in Python. They are also familiar to developers who’ve worked with libraries like Pydantic.

> As for the separate kwarg vs Classvar matter, this is really a design choice, we can discuss it in our next meeting.

I completely disagree that this is a subjective design choice. It's not about preference. It's about whether contributors can easily understand and use the tagging mechanism.  

> Either way, I think it's mostly a matter of documentation. We either have to tell users to add a class kwarg or a classvar, and in either case, it magically triggers a hidden part of the code, so it’s equally opaque to users.

I have to disagree here, too. This isn't just a matter of documentation. Ideally, contributors can look at a sum type in the codebase and immediately understand how to define their own.

- A class-level attribute like `kind: ClassVar[str] = "memmap"` is intuitive and self-explanatory for most Python developers.  
- A kwarg in a class definition (e.g., `type_="memmap"`) is not something anyone expects. It’s unconventional and requires explanation, which adds cognitive overhead.  

When you say “it's equally opaque,” I think you're discounting the value of familiarity. Familiar patterns reduce friction, while unconventional ones add it. By choosing the familiar approach, we make the tagging mechanism approachable for everyone.

> So the difference is really about the subjective matter of which one looks the scariest.

Yes, and I am here to tell you that the kwarg approach looks far scarier to most Python developers. Few people will have seen or used this pattern before, whereas tagging with class-level attributes is widely recognized and understood.

In summary, I strongly believe that the kwarg approach introduces unnecessary barriers to understanding and collaboration and should be replaced with class-level attributes before this PR is merged.

jlamypoirier:
I still have a slight preference for kwargs, but I'll switch it back to a class attribute to avoid an empty debate about what a typical python developer may or may not find understandable.

> 1. No contributor should need to understand the internal mechanics to define config classes. Whether it's `__init_subclass__` or not behind the scenes, it should be abstracted away. Nobody should need to dig into this unless they are maintaining that very core mechanism. Pydantic is a good example: nobody looks at its metaclass implementation, yet the interface is clear and consistent.

This won't be solved here either way, the only way to make the mechanics clear is by documenting it.

> 2. The tagging mechanism must be clear, conventional, and Pythonic. Using kwargs in class definitions isn't something any Python developers will expect. I've never seen this being used before, and I suspect others haven't either. Contributors need to understand how to define new configuration classes and tag them in a way that feels natural. Class-level attributes are well-known and widely used for tagging in Python. They are also familiar to developers who’ve worked with libraries like Pydantic.

Again highly subjective, but I won't argue further.

jlamypoirier:
> Alternatively, require subclasses to explicitly override the tag field, which can serve as both a type discriminator and a registration signal.

How would I do that? Other than implicitly by complaining on duplicates?

tscholak (approved):
Thanks @jlamypoirier, Lgtm

Issue #52, opened 2024-11-19
[feat] Support triton cross-entropy for larger vocabularies


jlamypoirier:
# 🧐 Problem Description

Vocab size is limited to 64K (I think?) because of triton's limitation on the block size. 

# 💡 Proposed Solution

The standard way is with looping over blocks, as is done for example in https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/ops/triton/cross_entropy.py (which was used as a basis for the Fast-LLM implementation).

 Fast-LLM got rid of looping to simplify things and optimize the case with a smaller vocab size, so we'll need to bring looping back. Some care will be needed to keep the current performance when looping is unnecessary (looping means multiple read of the logits, etc.).

# 🔄 Alternatives Considered

We have other implementations, but they are much slower.

# 📈 Potential Benefits

Faster training and lower memory usage with large vocab sizes.

# 📝 Additional Context

tscholak:
Can we use https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/ops/triton/cross_entropy.py in Fast-LLM directly?

jlamypoirier:
> Can we use https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/ops/triton/cross_entropy.py in Fast-LLM directly?

We're already using (a highly modified version of) it. I had removed the high vocab size part for performance, we need a way to add it back without sacrificinf performance.

PR #113, opened 2025-01-14
Allow loss masking for defined spans of characters

### LINKED ISSUE #109: [feat] Implement Loss Masking to Exclude Predefined Token Spans from LM Loss
### 🧐 **Problem Description**

In supervised fine-tuning (SFT) for decoder-only models, training data sequences typically include a mix of system prompts, user inputs, and model responses. However, the loss during fine-tuning should only be computed on the model responses. This practice, known as **loss masking**, has been shown to significantly improve model performance and generalization.

Currently, Fast-LLM does not support propagating token spans for loss masking from the dataset to the loss function, which limits its effectiveness for SFT. Implementing this feature requires:

1. A mechanism to identify and propagate spans corresponding to **excluded portions** (e.g., system prompts and user inputs).
2. Modifications to the training pipeline to exclude these spans during the loss computation.

---

### 💡 **Proposed Solution**

To support loss masking, we propose the following approaches, ranked by priority:

---

#### **1. Enhanced Memory-Mapped Indexed Format (Preferred Approach)**

- Extend the existing memory-mapped dataset format to include optional fields for token spans that specify which portions of the input should be excluded from the loss computation.
- Enhance the `prepare` command to:
  - Convert Hugging Face datasets into the memory-mapped format.
  - Extract and preprocess character spans into token spans during dataset preparation.
- **Assumptions**:
  - The `text` field contains fully formatted prompts (system prompt, user prompt(s), and model response(s)), optionally annotated with special chat tokens or tags.
  - Character spans align cleanly with token spans.

**Advantages**:
- Retains the scalability and efficiency of the memory-mapped dataset format.
- Integrates seamlessly into Fast-LLM’s current workflow.
- Supports both pretraining and SFT use cases.

**Trade-offs**:
- Adds complexity to the dataset preparation process.
- Assumes accurate character-to-token alignment, which may cause issues in edge cases.

---

#### **2. JSONL Dataset Format**

- Use a simpler dataset format with JSONL files, where each entry includes:
  - `"text"`: Fully formatted prompt containing system prompts, user inputs, and model responses.
  - `"masked_spans"`: A list of character span tuples to exclude from the loss computation.
- Create a dataset wrapper to read and process JSONL files during training.

**Example JSONL Entry**:
```json
{
    "text": "<system>\nYou are an assistant that provides concise and accurate answers.\n</system>\n<user>\nWhat is the capital of France?\n</user>\n<assistant>\nThe capital of France is Paris.\n</assistant>\n<user>\nCan you tell me more about Paris?\n</user>\n<assistant>\nParis is the capital city of France, known for its art, culture, and history.\n</assistant>",
    "masked_spans": [
        [0, 61],   // System message
        [62, 91],  // User prompt: "What is the capital of France?"
        [116, 146], // User prompt: "Can you tell me more about Paris?"
    ]
}
```

**Advantages**:
- Simplifies the pipeline, especially for smaller datasets.
- Avoids complex preprocessing and format conversions.

**Trade-offs**:
- Lacks scalability optimizations for larger datasets.
- Requires tight control over dataset formatting to avoid inconsistencies.

---

#### **3. Hugging Face Dataset Integration**

- Use Hugging Face datasets directly instead of the memory-mapped format.
- Perform tokenization on-the-fly and dynamically extract spans for loss masking.
- Require users to specify a unified chat format to delineate system prompts, user inputs, and model responses.

**Advantages**:
- User-friendly and works well with smaller datasets already in Hugging Face format.
- Reduces the need for format conversions.

**Trade-offs**:
- On-the-fly tokenization introduces runtime overhead.
- Not optimized for large-scale pretraining datasets.

---

### **Pipeline Integration**

For all approaches, token spans for excluded portions of the input should be passed as part of the `kwargs` in the model’s forward function:
```python
def forward(self, input_: torch.Tensor, kwargs: dict):
    masked_spans = kwargs.pop("masked_spans", [])
    ...
```

This masking logic ensures that the loss computation is limited to the relevant portions (e.g., model responses).

---

### 🔄 **Alternatives Considered**

- **Ignore Loss Masking**: Simplifies implementation but results in suboptimal performance.
- **Invert the Mask**: Specify spans to **include** in the loss computation instead of excluding spans. However, this approach risks producing NaN losses if no spans are provided.

---

### 📈 **Potential Benefits**

- **Improved Model Performance**: Aligns loss computation with best practices in SFT, enhancing both performance and generalization.
- **Flexibility**: Supports multiple dataset formats for diverse user needs.
- **Ease of Use**: Automates span extraction, reducing user overhead.

---

### 📝 **Additional Context**

- This feature is critical for SFT.
- Reference PR for RLHF GRPO: [#20](https://github.com/ServiceNow/Fast-LLM/pull/20).

---

### **Deliverables**

1. **Code Implementation**:
   - Extend dataset handling and pipeline logic for one or more proposed formats.
   - Modify the loss computation to respect token spans for loss masking.
   - Add tests.
2. **Documentation**:
   - Add clear guidelines for preparing datasets for loss masking.
   - Provide examples for each supported format.

sohamparikh:
# ✨ Description
Support loss masking for spans specified in the input data. This PR will ensure that loss will not be computed on the specified spans. The biggest use-case for this is instruction tuning data where we want to avoid training on the prompts. 

Closes #109 

## 📝 Changes

- [x] Support character spans as inputs specified in the prepare command
- [x] Read the spans during training and apply masks to cross-entropy loss

## 🔍 Type of change

---


fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -92,20 +114,23 @@ def get_document_sizes(self) -> "np.ndarray":
         return self._document_sizes
 
     @classmethod
-    def write_dataset(cls, prefix: pathlib.Path | str, documents: typing.Iterable[np.ndarray]):
+    def write_dataset(cls, prefix: pathlib.Path | str, documents: typing.Iterable[tuple[np.ndarray, np.ndarray]]):
```
tscholak, line 117:
```suggestion
    def write_dataset(cls, prefix: pathlib.Path | str, documents: typing.Iterable[Document]):
```
where
```
@dataclasses.dataclass
class Document:
  text: np.ndarray
  mask_spans: np.ndarray
```

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -48,6 +48,25 @@ def _init(self, name: str, prefix: pathlib.Path | str):
             offset=offset + self._document_sizes.nbytes,
         )
 
+        self._num_spans = np.frombuffer(
```
tscholak, line 51: can you make this backwards-compatible, so that older data can still be loaded that doesn't have spans?

jlamypoirier:
Looks good so far, but can you please add a short description and/or point to an issue?

fast_llm/data/dataset/gpt/memmap.py
sohamparikh: is it ok to bump the version for that? seems to be the cleanest approach
`idx_stream.write(struct.pack("<Q", 2))  # Version`

tscholak: clever idea, yes!

fast_llm/data/preparator/gpt_memmap/prepare.py
```diff
@@ -22,14 +22,48 @@ class GPTMemmapDatasetPreparator[ConfigType: GPTMemmapDatasetPreparatorConfig](D
...
+        Perform span-aware tokenization and return the tokenized input_ids along with token spans.
+        """
+        input_ids = []
+        token_spans = []
+        char_pos = 0
+        beginning_of_text = True
+        for start, end in char_spans:
+            if char_pos < start:
+                curr_text = text[char_pos:start]
+                tokenized_text = self._tokenizer.tokenize(curr_text, add_special_tokens=beginning_of_text)
```
tscholak, line 36: This works only for those tokenizers that only have a BOS but not a EOS token.
For those that come with both, can we control whether `tokenize` adds the BOS and EOS tokens independently? I'm worried that we are adding the EOS token at the end of the first segment and the BOS token at the beginning of the last segment.

fast_llm/functional/cross_entropy.py
```diff
@@ -96,14 +111,15 @@ def parallel_cross_entropy_forward_backward(
...
         )
         exp_logits2 = exp_logits1.mul((grad_output / logits.size(0)) / sum_exp_logits.unsqueeze(dim=-1))
         if logits_scale_factor != 1.0:
             exp_logits2 *= logits_scale_factor
 
-        grad = exp_logits2.to(logits.dtype)
+        grad.index_put_((mask,), exp_logits2.to(logits.dtype))
 
     predicted_logits = (target_mask * logits_norm.gather(1, target)).squeeze(1)
     all_reduce(predicted_logits, op=ReduceOp.SUM, group=group)
```
tscholak, line 125: does triton support masking?

fast_llm/data/preparator/gpt_memmap/prepare.py
sohamparikh: good catch! I'll make it explicitly add BOS only for the first segment
Btw, most tokenizers (`Llama-3.1`, `Mistral-Nemo-Base-2407`, `OLMoE-1B-7B-0924`) do not add the EOS token with `add_special_tokens=True`. Does this mean we've been training the models without the EOS token?

In the future I think we should make this config driven. The default behaviour would be to add both BOS and EOS tokens. It's important for pretraining with attention mask, and especially for SFT.

tscholak: > Does this mean we've been training the models without the EOS token?

Indeed, we decided that adding both BOS and EOS tokens in pretraining was unnecessary, because they are redundant. Here though I think we need to add them because we need to teach the model to terminate a response with the EOS token so that generation can stop at the right moment. Btw, I think HF is not adding the EOS token by default because otherwise prompts would end with it.

fast_llm/functional/cross_entropy.py
tscholak: I think it doesn't: https://github.com/ServiceNow/Fast-LLM/blob/soham/loss-masking-spans/fast_llm/functional/triton/cross_entropy.py
We need to add it. Since this is the same for all loss functions, it would make sense to implement it before dispatching to specialized cross-entropy implementations:

```python
def cross_entropy_forward_backward(
    logits,
    target,
    grad_output: float | None,
    group: ProcessGroup | None,
    implementation: CrossEntropyImpl = CrossEntropyImpl.fused,
    logits_scale_factor: float = 1.0,
    ignore_index: int=-100,
) -> tuple[torch.Tensor, torch.Tensor | None]:
    ...
    mask = target != ignore_index
    target = target[mask]
    logits = logits[mask]
    ...
```

fast_llm/data/data/gpt/data.py
```diff
@@ -23,6 +26,19 @@
...
+@dataclasses.dataclass
+class GPTDataBatch:
+    ids: torch.Tensor
+    spans: torch.Tensor
+
+
+def gpt_data_collate_fn(batch: list[GPTSample]) -> GPTDataBatch:
+    stacked_ids = np.stack([sample.ids for sample in batch])
+    # stacked_spans = np.stack([sample.spans for sample in batch])
+    stacked_spans = [torch.from_numpy(sample.spans) for sample in batch]
```
tscholak, line 38: possible type error. that's a list of tensors and not a tensor, isn't it?

fast_llm/data/dataset/gpt/fim.py
```diff
@@ -57,26 +58,26 @@ def _fim(self, sample: np.ndarray, np_rng: np.random.RandomState) -> np.ndarray:
...
         if diff > 0:  # too long
-            sample = sample[:sample_len]
+            sample.ids = sample.ids[:sample_len]
         elif diff < 0:  # too short
-            sample = np.concatenate([sample, np.full((-1 * diff), self._pad_tok_id)])
+            sample.ids = np.concatenate([sample.ids, np.full((-1 * diff), self._pad_tok_id)])
 
-        assert sample.shape[0] == sample_len
+        assert sample.ids.shape[0] == sample_len
         return sample
```
tscholak, line 81: since this code will change the order of tokens in the sequence, we would need to change the masks accordingly to allow for FIM with loss masking.
At this point, I think we should not and fail if FIM was used with loss masking.

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -65,13 +105,19 @@ def __del__(self):
...
+        ids = np.frombuffer(
             self._bin_buffer,
             dtype=self._dtype,
             count=self._document_sizes[idx] - offset if length is None else length,
             offset=self._pointers[idx] + offset * np.dtype(self._dtype).itemsize,
         )
+        spans = []
+        for span in self._spans[idx]:
+            if span[0] < offset + len(ids) and span[1] >= offset:
+                spans.append([max(span[0], offset) - offset, min(span[1], offset + len(ids) - 1) - offset])
```
tscholak, line 118: this code should also use `length` if it is `not None`

fast_llm/data/dataset/gpt/random.py
```diff
@@ -31,10 +32,22 @@ def __init__(self, config: GPTSamplingConfig, name: str):
...
+        spans = []
+        prev_end = -1
+        for _ in range(n_spans):
+            if prev_end >= len(ids) - 1:
+                break
+            start = np.random.RandomState(np_seed).randint(prev_end + 1, len(ids))
+            end = np.random.RandomState(np_seed).randint(start, len(ids))
+            spans.append([start, end])
+            prev_end = end
+        return GPTSample(ids=ids, spans=np.array(spans, dtype=np.int32).reshape(-1, 2))
```
tscholak, line 50: that's a nice addition, though I'm not sure if the random dataset actually needs spans... We use this only for benchmarking purposes to measure training performance without being IO bound.

fast_llm/data/tokenizer.py
```diff
@@ -33,8 +35,52 @@ def vocab(self) -> dict[str, int]:
...
+            beginning_of_text = False
+            token_spans.append((len(input_ids), len(input_ids) + len(tokenized_text) - 1))
+            input_ids.extend(tokenized_text)
+            char_pos = end + 1
+        if char_pos < len(text):
+            curr_text = text[char_pos:]
+            tokenized_text = self.tokenize(curr_text)
+            input_ids.extend(tokenized_text)
+        if self.special_tokens_mode in [SpecialTokensMode.eos_only, SpecialTokensMode.bos_eos]:
+            input_ids.append(self.eod_id)
```
tscholak, line 82: shouldn't this be handled by `self.tokenize`?

fast_llm/functional/cross_entropy.py
```diff
@@ -51,14 +61,15 @@ def fused_cross_entropy_forward_backward(
     if grad_output is None:
         grad = None
     else:
+        grad = torch.zeros((loss_mask.size(0), *logits.shape[1:]), dtype=logits.dtype, device=logits.device)
```
tscholak, line 64: can you put this and the `index_put_` below also in the dispatch function?

fast_llm/data/dataset/gpt/random.py
sohamparikh: I added it for the tests. Can make the spans empty if you think it'll mess with the benchmark numbers

fast_llm/data/dataset/gpt/memmap.py
sohamparikh: the returned spans are indexed for `ids` which already take into account `length` and are 0-indexed starting from `offset`. I don't see a situation where `length` is needed, am I missing something?

tscholak: oh, you're right then!

tscholak (commented):
Hi @sohamparikh, nice progress!
Is anything functional missing from this PR? Do the spans make it all the way to the loss function, and does packing work as expected with spans? Can we test this?

sohamparikh:
Functionally it's good to go now.
I've tested `prepare` and `train` on SmolLM2-135M on a single GPU using a dummy dataset with spans. Seems to be working fine, including packing and the loss functions.

What kind of tests would you like to see? I can test it on a bigger model with multiple nodes if that makes sense

fast_llm/data/dataset/gpt/fim.py
sohamparikh: throwing an error statement in this function now

fast_llm/functional/cross_entropy.py
```diff
@@ -126,19 +136,29 @@ def cross_entropy_forward_backward(
...
-        return parallel_cross_entropy_forward_backward(
+        loss, grad_logits = parallel_cross_entropy_forward_backward(
             logits, target, grad_output, group, logits_scale_factor=logits_scale_factor
         )
     else:
-        return _CROSS_ENTROPY_IMPLEMENTATIONS[implementation](
+        loss, grad_logits = _CROSS_ENTROPY_IMPLEMENTATIONS[implementation](
             logits, target, grad_output, logits_scale_factor=logits_scale_factor
         )
+    if grad_logits is not None:
```
jlamypoirier, line 159: This needs to go inside the implementation because each of them can be optimized in its own way. torch implementation has `ignore_index` already, compiled version can include this inside the compile block, and triton kernels can include masking. For the triton part you can keep this if you don't know triton (it's a really easy one though).

Also `torch.where` would do a better job here.

(And as usual, masking needs to be opt-in)

fast_llm/data/data/gpt/data.py
```diff
@@ -23,6 +26,19 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass
+class GPTDataBatch:
```
jlamypoirier, line 30: `GPTBatch` is enough

tests/test_dataset.py
```diff
@@ -82,8 +83,8 @@ def get_test_data_and_samples(
     batch_config.setup(distributed_config)
     batch_config.validate()
     samples = {
-        phase: [batch[0] for batch in data.get_iterator(batch_config, phase, consumed_samples=0, num_workers=0)]
-        for phase, samples in samples_per_phase.items()
+        phase: list(data.get_iterator(batch_config, phase, consumed_samples=consumed_samples, num_workers=0))
```
jlamypoirier, line 86: This makes the existing tests too complicated. Instead, Please test spans with a small number of separate tests specifically targeting them. (Not sure we need full coverage for all cases, or you could make one complicated test case that indirectly test many classes.)

fast_llm/data/dataset/gpt/indexed.py
```diff
@@ -36,6 +36,9 @@ def get_document_sizes(self) -> np.ndarray:
         # TODO: This can be really big.
         return self._dataset.get_document_sizes()[self._begin : self._end]
 
+    def get_span_sizes(self) -> np.ndarray:
```
jlamypoirier, line 39: Is this only used in the tests? If so I'm not sure it's worth making a public method at this stage.
(And would need to be added to GPTIndexedDataset too)

jlamypoirier (changes requested):
I reviewed the overall structure, looks good for the intended purpose but will need polishing. Main areas to focus on (see individual comments):
* Spans need to be opt-in, so that there is a negligible impact when they are not used (which is most of the time).
* Turning samples and friends into dataclasses is a good idea but goes a bit outside the present scope. Ideally it would go in a separate PR, but it's ok to include here if done well (see coments)
* Names need to follow our style guide (https://servicenow.github.io/Fast-LLM/contributing/style-guide/) a bit better. Please use self-descriptive names as much as possible, ex. `spans` is a bit cryptic and could mean lots of things. (Not sure what to replace it with, but it should hint to loss masking in some way)

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -92,14 +138,25 @@ def get_document_sizes(self) -> np.ndarray:
...
+        """
+        The number of spans in each document in the dataset.
+        The resulting array could be very large, so this method should be called cautiously,
+        and derived classes should try to avoid holding the whole array im memory.
+        """
+        return self._num_spans
+
     @classmethod
-    def write_dataset(cls, prefix: pathlib.Path | str, documents: typing.Iterable[np.ndarray]):
+    def write_dataset(cls, prefix: pathlib.Path | str, documents: typing.Iterable[GPTMemmapDocument]):
```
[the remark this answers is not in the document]
sohamparikh: Adding a byte would make it incompatible with already tokenized datasets. 

If we want to keep the spans opt-in, I can keep the version as 1, move `num_spans` and `spans` to the end of `.idx` file (based on a flag) and in `_init`, we expect a flag (from some config, not sure which) indicating that we want to read spans.


@tscholak's comment for reference: https://github.com/ServiceNow/Fast-LLM/pull/113#discussion_r1916731903

fast_llm/functional/cross_entropy.py
sohamparikh: > (And as usual, masking needs to be opt-in)

Do you mean an additional flag indicating whether loss masking should take place (using the config option for reading spans)? 
I'm not clear why `ignore_index` isn't sufficient since it wouldn't be set without the spans config flag anyway

jlamypoirier: Yes. `ignore_index` isn't sufficient because it would slow things down when not in use.

fast_llm/data/dataset/gpt/memmap.py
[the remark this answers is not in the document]
jlamypoirier: There is no problem with the extra byte because we're bumping the version. The older datasets have version 1 so we don't read it.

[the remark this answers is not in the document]
sohamparikh: Got it, I misunderstood your first comment. This makes sense to me now

sohamparikh:
Hi @jlamypoirier,
I've made these changes based on your review

1. Spans and loss-masking are opt-in, both during `prepare` and `train`. During `prepare`, spans will be included if `spans_field` is specified, otherwise tokenization and dataset storage will be the same as before (only the version is bumped to 2). During `train`, we read the spans only if `use_loss_masking_spans` is specified in the data config, else we ignore them i.e., same as previous behavior.
2. Cross-entropy implementations will enable loss-masking only if `apply_loss_mask` is True in the model config, else they will continue according to their flow. I believe the triton implementation already applies the loss mask for -ve IDs, so I just added an extra check for `ignore_index` by default. Lmk if you prefer that we mask `ignore_index` iff `apply_loss_mask=True`
3. Reduced the dataclasses to `GPTSample` and `GPTBatch`. I'll create an issue for creating a `Sample` (and `Batch`?) class which other classes inherit from. `loss_masking_spans` are now None by default, and more in line with being opt-in.
4. Reverted original tests to check for `token_ids` only, and added a basic test for spans

fast_llm/data/data/gpt/data.py
```diff
@@ -23,6 +26,20 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass
+class GPTBatch:
+    token_ids: torch.Tensor
+    loss_masking_spans: list[torch.Tensor]
```
tscholak, line 32:
```suggestion
    loss_masking_spans: list[torch.Tensor] | None
```

fast_llm/data/tokenizer.py
```diff
@@ -33,8 +35,55 @@ def vocab(self) -> dict[str, int]:
     def inv_vocab(self) -> dict[int, str]:
         return self._inv_vocab
 
-    def tokenize(self, text: str) -> list[int]:
-        return self.tokenizer.encode(text)
+    def tokenize(self, text: str, beginning_of_text=True, end_of_text=True) -> list[int]:
```
jlamypoirier, line 38: Can be shortened to `begin`, `end`

fast_llm/data/dataset/gpt/fim.py
```diff
@@ -150,9 +153,11 @@ def _fim_permute_sequence(
         middle = contents[boundaries[0] : boundaries[1]]
         suffix = contents[boundaries[1] :]
 
-        prefix = np.array([*self._tokenizer.tokenize(prefix)], dtype=np.int64)
-        middle = np.array([*self._tokenizer.tokenize(middle)], dtype=np.int64)
-        suffix = np.array([*self._tokenizer.tokenize(suffix)], dtype=np.int64)
+        prefix = np.array([*self._tokenizer.tokenize(prefix, end_of_text=False)], dtype=np.int64)
```
jlamypoirier, line 156: Does that mean FIM wasn't working before?

jlamypoirier (commented):
A few more comments, mainly about sorting out `use_loss_masking_spans`, tokenization and cross-entropy

fast_llm/functional/cross_entropy.py
```diff
@@ -58,7 +72,10 @@ def fused_cross_entropy_forward_backward(
         if logits_scale_factor != 1.0:
             exp_logits *= logits_scale_factor
 
-        grad = exp_logits.to(logits.dtype)
+        if apply_loss_mask:
+            grad = torch.where(loss_mask.unsqueeze(1), exp_logits.to(logits.dtype), 0)
```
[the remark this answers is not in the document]
sohamparikh: thanks for catching this!

fast_llm/data/tokenizer.py
```diff
@@ -33,8 +35,55 @@ def vocab(self) -> dict[str, int]:
...
+        if self.sequence_delimiters == SequenceDelimiters.eos_only:
```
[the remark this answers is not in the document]
tscholak: Ok, let's just do bos-eos and nothing else. cc @sohamparikh

fast_llm/data/tokenizer.py
```diff
@@ -16,9 +16,11 @@ def __init__(self, config: TokenizerConfig):
         self.tokenizer: PreTrainedTokenizerFast = PreTrainedTokenizerFast.from_pretrained(
             pretrained_model_name_or_path=config.path, errors="replace", max_len=None
         )
+        self.sequence_delimiters = config.sequence_delimiters
         if self.tokenizer.eos_token_id is None:
             raise ValueError("Tokenizer does not have an EOS token.")
         self.eod_id = self.tokenizer.eos_token_id
+        self.bod_id = self.tokenizer.bos_token_id
```
tscholak, line 23:
```suggestion
        if self.tokenizer.bos_token_id is None:
            raise ValueError("Tokenizer does not have an BOS token.")
        self.bod_id = self.tokenizer.bos_token_id
```
add assertion

fast_llm/data/dataset/gpt/fim.py
sohamparikh: Depends on the tokenizer. Starcoder tokenizer does not add any special tokens, so it was fine for that. Llama/Mistral add bos, so probably not if fim was used with that

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -48,6 +53,27 @@ def _init(self, name: str, prefix: pathlib.Path | str) -> None:
...
+            self._num_spans = np.frombuffer(
+                self._index_bin_buffer,
+                dtype=np.int32,
+                count=self._num_documents,
+                offset=offset + self._document_sizes.nbytes + self._pointers.nbytes,
+            )
+            span_offset = offset + self._document_sizes.nbytes + self._pointers.nbytes + self._num_spans.nbytes
+            self._num_spans_cumsum = np.r_[0, np.cumsum(self._num_spans[:-1], dtype=np.int64)]
+            for idx in range(self._num_documents):
+                self._spans.append(
```
tscholak, line 68: can we skip unpacking the spans if `use_loss_masking_spans` is turned off?

fast_llm/data/data/gpt/data.py
```diff
@@ -23,6 +26,20 @@
...
+    token_ids: torch.Tensor
+    loss_masking_spans: list[torch.Tensor] | None
+
+
+def gpt_data_collate_fn(batch: list[GPTSample]) -> GPTBatch:
+    stacked_ids = np.stack([sample.token_ids for sample in batch])
+    stacked_spans = None
+    if batch[0].loss_masking_spans is not None:
+        stacked_spans = [torch.from_numpy(sample.loss_masking_spans) for sample in batch]
+    return GPTBatch(token_ids=torch.from_numpy(stacked_ids), loss_masking_spans=stacked_spans)
```
tscholak, line 40:
```suggestion
def gpt_data_collate_fn(batch: list[GPTSample], use_loss_masking_spans: bool) -> GPTBatch:
    stacked_ids = np.stack([sample.token_ids for sample in batch])
    stacked_spans = None
    if use_loss_masking_spans:
        stacked_spans = [torch.from_numpy(sample.loss_masking_spans) for sample in batch]
    return GPTBatch(token_ids=torch.from_numpy(stacked_ids), loss_masking_spans=None)
```
is that possible?
what does `loss_masking_spans = None` mean? usage of loss masking spans turned off globally?

fast_llm/data/data/gpt/data.py
```diff
@@ -120,6 +138,7 @@ def get_iterator(
                 num_workers=num_workers,
                 prefetch_factor=prefetch_factor,
                 pin_memory=True,
+                collate_fn=gpt_data_collate_fn,
```
tscholak, line 141:
```suggestion
                collate_fn=partial(gpt_data_collate_fn, use_loss_masking_spans=use_loss_masking_spans),
```

fast_llm/data/data/gpt/data.py
sohamparikh: yes, that was the hacky way of figuring it out. Your suggestion is much cleaner

fast_llm/data/tokenizer.py
```diff
@@ -33,8 +35,43 @@ def vocab(self) -> dict[str, int]:
...
         return self._inv_vocab
 
-    def tokenize(self, text: str) -> list[int]:
-        return self.tokenizer.encode(text)
+    def tokenize(self, text: str, begin=True, end=True) -> list[int]:
+        return (
+            ([self.bod_id] if begin else [])
+            + self.tokenizer.encode(text, add_special_tokens=False)
+            + ([self.eod_id] if end else [])
+        )
```
tscholak, line 43: I think we can delegate to Huggingface's `build_inputs_with_special_tokens` function. Here's a couple of example implementations:

* https://github.com/huggingface/transformers/blob/0de15c988b0d27758ce360adb2627e9ea99e91b3/src/transformers/models/gemma/tokenization_gemma.py#L222

* https://github.com/huggingface/transformers/blob/0de15c988b0d27758ce360adb2627e9ea99e91b3/src/transformers/models/llama/tokenization_llama_fast.py#L246

Behaviour seems to be the same for most decoder-only models. I therefore think we can just set `self.tokenizer.add_eos_token` and `self.tokenizer.add_bos_token` to `True` or `False` and get the desired outcome. It's a bit iffy because these settings are stateful.

sohamparikh: It's not true for all tokenizers. By default, they use [PretrainedTokenizerBase.build_inputs_with_special_tokens](https://github.com/huggingface/transformers/blob/0de15c988b0d27758ce360adb2627e9ea99e91b3/src/transformers/tokenization_utils_base.py#L3419C5-L3419C41) which does not handle `bos` or `eos`
For example, `Mistral-Nemo-Base-2407` uses `PretrainedTokenizerFast` and falls back to using `PretrainedTokenizerBase.build_inputs_with_special_tokens`

fast_llm/data/dataset/gpt/memmap.py
sohamparikh: I think @jlamypoirier [suggested always reading them](https://github.com/ServiceNow/Fast-LLM/pull/113#discussion_r1941929171) to avoid passing unnecessarily passing `use_loss_masking_spans`?

fast_llm/data/data/gpt/data.py
```diff
@@ -23,6 +27,20 @@
...
+    token_ids: torch.Tensor
+    loss_masking_spans: list[torch.Tensor] | None
+
+
+def gpt_data_collate_fn(batch: list[GPTSample], use_loss_masking_spans: bool) -> GPTBatch:
+    stacked_ids = np.stack([sample.token_ids for sample in batch])
+    stacked_spans = None
+    if use_loss_masking_spans:
+        stacked_spans = [torch.from_numpy(sample.loss_masking_spans) for sample in batch]
+    return GPTBatch(token_ids=torch.from_numpy(stacked_ids), loss_masking_spans=None)
```
tscholak, line 41:
```suggestion
    return GPTBatch(token_ids=torch.from_numpy(stacked_ids), loss_masking_spans= stacked_spans)
```
sorry I gave you wrong code earlier :/

fast_llm/data/tokenizer.py
tscholak: oh, ok then

fast_llm/functional/cross_entropy.py
```diff
@@ -58,23 +72,36 @@ def fused_cross_entropy_forward_backward(
...
+    apply_loss_mask: bool = False,
 ) -> tuple[torch.Tensor, torch.Tensor | None]:
     """
     A fused implementation of cross-entropy with torch compile, with support for tensor parallelism.
     Comes with a noticeable overhead, but reduces memory usage.
     """
     # TODO: Compiled version incorrect for some inputs (32 bit indexing issue?).
     # TODO: Optimize, overlap/combine reductions
+    if apply_loss_mask:
+        loss_mask = target != ignore_index
```
tscholak, line 104: do you also need the clamp here like above?

fast_llm/functional/cross_entropy.py
```diff
@@ -9,7 +9,12 @@
 
 
 def torch_cross_entropy_forward_backward(
-    logits: torch.Tensor, target: torch.Tensor, grad_output: float | None, logits_scale_factor: float = 1.0
+    logits: torch.Tensor,
+    target: torch.Tensor,
+    grad_output: float | None,
+    logits_scale_factor: float = 1.0,
+    ignore_index: int = -100,
+    apply_loss_mask: bool = False,
```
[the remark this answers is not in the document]
sohampnow: `torch.nn.functional.cross_entropy` already takes care of `ignore_index`. We can remove `apply_loss_mask`, but I kept it for simplicity in the dispatch function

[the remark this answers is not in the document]
sohamparikh: `torch.nn.functional.cross_entropy` already takes care of ignore_index. We can remove `apply_loss_mask` from args, but I kept it for simplicity in the dispatch function

fast_llm/functional/cross_entropy.py
sohamparikh: lines 119-121 take care of it with masking (instead of clamping)

fast_llm/functional/cross_entropy.py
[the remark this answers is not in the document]
tscholak: But isn’t the contract here that if apply loss mask is false, then the target labels equal to ignore index are ignored?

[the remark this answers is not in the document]
sohamparikh: when `use_loss_masking_spans=True`, we hardcode `ignore_index=-100` which will be masked by `torch.nn.functional.cross_entropy` by default.

We can have this logic, but it'll still mask losses when `use_loss_masking_spans=True` and `apply_loss_mask=False`

```
if apply_loss_mask:
    loss = torch.nn.functional.cross_entropy(logits_, target, ignore_index=ignore_index).mean()
else:
    loss = torch.nn.functional.cross_entropy(logits_, target).mean()
```

We can also change the default value of `ignore_index` in Fast-LLM to make it honor the contract (in most cases)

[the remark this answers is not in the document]
tscholak: I see, what you're saying is that `torch.nn.functional.cross_entropy` cannot be prevented from masking. If any target labels are equal to ignore index, `torch.nn.functional.cross_entropy` will mask them.

fast_llm/functional/cross_entropy.py
```diff
@@ -41,6 +51,10 @@ def fused_cross_entropy_forward_backward(
     """
     # Do the forward and backward passes all at once, and fused with dtype conversion.
     # Way faster and more memory-efficient than the pytorch version.
+    if apply_loss_mask:
+        loss_mask = target != ignore_index
```
tscholak, line 55: why are we first integrating the loss mask with the target tensor and then extract it here again? does this save any noteworthy amount of memory? why not passing an explicit and optional binary loss mask to the cross entropy function?

tscholak: cc @jlamypoirier

fast_llm/functional/cross_entropy.py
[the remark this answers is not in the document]
sohamparikh: correct

tests/data/test_fim.py
```diff
@@ -13,12 +13,12 @@
...
-    [86, 89, 22255, 1073, 79, 480],
+    [86, 49152, 89, 22255, 1073, 79],
     [8008, 498, 71, 727, 80, 315],
     [2210, 8179, 73, 2582, 897, 1178],
-    [86, 89, 88, 87, 409, 70],
-    [86, 83, 744, 89, 64, 333],
-    [86, 89, 1461, 87, 330, 7876],
+    [86, 89, 88, 49152, 87, 49152],
+    [86, 49152, 83, 744, 89, 64],
+    [86, 89, 1461, 49152, 87, 49152],
```
sohamparikh, line 21: @jlamypoirier can you pls check if this is good with you?
I'm not sure if this is ok for FIM.

It changed since we're now specifying `begin` and `end` in the tokenizer. The test tokenizer doesn't add either by default so we didn't see `49152` earlier. It is both `bod_id` and `eod_id` for that tokenizer

fast_llm/functional/cross_entropy.py
tscholak: I don't want to delay this PR unnecessarily. If you and @jlamypoirier think that the combo of ignore indices in the `target` and `apply_loss_mask` is the right interface for the cross entropy loss, then let's leave it like that.

jlamypoirier (approved):
Did a few small changes, should be good to merge if nobody opposes:
* Simplified the cross-entropy thing. I had a deeper look and determined that the best option is to standardize all methods and always enable masking for index < 0, which makes all the `apply_loss_mask` arguments unnecessary.. The performance impact won't be a real problem, it was already on for triton which is the one we care about the most, and for the compiled ones the overhead will probably be made negligible with compilation. Only remaining issue is torch cross-entropy having the different behaviour of masking -100 only.
* Simplified the tests
* Replaced the random spans in tests with a much faster method
* Fixed some other tests.

tscholak:
@sohamparikh, can you run two experiments with SFT data, one with masking turned on and one with masking turned off? I'd like us to see what the performance impact of masking is, if any.

tscholak:
ok, so 13% performance loss with masking turned on.
Are we happy with that? Can we expect better? @jlamypoirier?
I was hoping for virtually no penalty, but not sure how realistic that is.

sohamparikh:
image
with `cross_entropy_impl: triton` they're pretty similar

sohamparikh:
@tscholak pls disregard my plot for fused cross-entropy above, I've deleted it. Found a minor bug in fused (and the previous runs are unreliable). Will re-run and post again

tscholak:
> with cross_entropy_impl: triton they're pretty similar (smaller vocab size this time)

smaller vocab because of #52? how small did you go? We are forced to use the fused implementation for our models right now, and the penalty there is 13% according to your earlier plot?

jlamypoirier:
@tscholak Can you  post more details about the 13%? I set the masking to be always on, so that looks really bad

sohamparikh:
Pls allow me to clear the confusion I created with the multiple (now deleted) plots. The numbers are generally quite similar both with and without loss-masking irrespective of the cross-entropy kernel used.
However, I do see that the first run is (almost) always slower than the second with the exact same config (even with a different run directory, and hence a different dataset cache).

| Model | Loss Mask | Cross-Entropy | Vocab Size | Run 1 (tok/s/gpu) | Run 2 (tok/s/gpu) |
|-------|-----------|---------------|------------|--------------|--------------|
| SLAM-5.1B | No | Triton | 49k | 18700 | 18700 |
| SLAM-5.1B | Yes | Triton | 49k | 16200 | 18700 |
| SLAM-5.1B | No | Fused | 49k | 16200 | 18650 |
| SLAM-5.1B | Yes | Fused | 49k | 16200 | 18700 |
| SLAM-5.1B | No | Triton | 131k | Crash |Crash |
| SLAM-5.1B | Yes | Triton | 131k | Crash | Crash |
| SLAM-5.1B | No | Fused | 131k | 14900 | 17400 |
| SLAM-5.1B | Yes | Fused | 131k | 14900 |  17350 |


So, good news: no 13% penalty

All the runs are in this wandb project:
https://wandb.ai/akshaykalkunte/slam

jlamypoirier:
Ok looks good then. The first run looks like a data loader bottleneck, second has it in local filesystem cache so that's normal.  We don't care much about masked vs not masked for the model since it's running the same code now. What we care about is the possible regression for fused without loss masking, and the lack of difference with triton seem to show it's not significant.

tscholak:
In this cases shouldn't we compare with `main`?

jlamypoirier:
> In this cases shouldn't we compare with `main`?

Ideally yes, but the triton implementation is unchanged so we can deduce the performance in `main`

sohamparikh:
Confirmed that the numbers with main are also similar.
https://wandb.ai/akshaykalkunte/slam/runs/ohq71uk8v1btd8sc

I'll merge this now

PR #129, opened 2025-01-24
Better timeouts

jlamypoirier:
# ✨ Description

Fix:

Set a custom timeout for slow operations (data setup, checkpoint save/load/export) using the well-hidden add_ephemeral method. Using a single "long timeout" for simplicity, the alternative would be to have 4-5 separate timeouts, which would be more powerful and actually simpler to implement but also more annoying to configure.

## 🔍 Type of change

Select all that apply:

### DIFF
--- a/fast_llm/data/dataset/gpt/config.py
+++ b/fast_llm/data/dataset/gpt/config.py
@@ -2,6 +2,7 @@ import dataclasses
 import enum
 import json
 import pathlib
+import time
 import typing
 
 from fast_llm.config import Config, Field, FieldHint, FieldUpdate, check_field, config_class, skip_valid_if_none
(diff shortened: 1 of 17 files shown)


fast_llm/data/dataset/gpt/config.py
```diff
@@ -364,3 +365,25 @@ def build_and_sample(self, config: GPTSamplingConfig) -> SampledDataset:
                 )
 
         return dataset_config.build_and_sample(config)
+
+
+@config_class()
+class GPTTestSlowDatasetConfig(GPTSampledDatasetConfig):
```
tscholak, line 371: why can't this be moved to a module in the tests?

jlamypoirier: Because it has to be run in a separate process (for distributed) which doesn't import the test file.

tscholak: I see, so you're calling the fast-llm CLI in a subprocess in the tests.

Issue #132, opened 2025-01-28
[feat] Speed up dataset sampling


jlamypoirier:
# 🧐 Problem Description

Dataset sampling takes a lot of time at the beginning of training. This means slower starts, which is bad for productivity, on top of causing trouble like timeouts (#76). 

# 💡 Proposed Solution

For blending, we can completely eliminate pre-sampling and instead sample on the fly, so let's do that (#133).

Indexed dataset sampling is more complicated, I don't think it's possible to completely remove pre-sampling, but we can still do a lot. 

**Shuffling is by far the biggest problem** (see benchmark below). Not only it takes most of the time, but it also forces oversampling when we train for less than one epoch. Potential approaches:
1. Skip shuffling entirely (see also #130). This is the appropriate approach for pre-shuffled datasets that run for less than one epoch, and for their first epoch if there is more than one, but won't do in general.
2. Shuffle with GPU. We can use torch.randperm, also worth benchmarking on cpu.Not sure about resulting quality, but definitely won't reproduce numpy results exactly.
3. Shuffle in parallel: we can parallelize over epochs if we shuffle them independently (see #130). Might help, but hard to implement and useless for single-epoch datasets.

I'd opt for 2, supplemented by 1 when applicable.

The distant second is the **saving of indexes**. We can do without optimizing it, but the arrays can get huge so it's also a disk space optimization. The doc index, is typically the biggest, especially is the sequences are small. Some options:
1. Again, avoiding shuffling would help. Without it, both the doc and shuffle indexes are unnecessary, dropping the vast majority of the saved data.
2. Recompute indexes on the fly, one epoch at the time. Possible for small datasets with independently shuffled datasets for which we only shuffle small chunks, but not for bigger ones (use too much memory, lag spikes at recomputation).

I'd opt for 2 when it makes sense (ex. with a configurable threshold), supplemented by 1 when applicable.

Then comes the **actual sampling using the c++ code**. This is also optional, but it would be good to drop the c++ plugin entirely. I was able to make a prototype that samples on the fly in #110, but it comes at a big cost: We need to pre-compute the sample starting points in the doc index, aka `document_sizes[doc_idx].cumsum`, and it takes twice the size of the doc index (because it's in int64). So don't really gain anything in the end. This could be fixed by saving sparse values in the array (ex. 1 in 10, or 1 in 100), for some extra work in the data loader (probably ok).

[...]

PR #133, opened 2025-01-28
Speedup dataset blending

jlamypoirier:
# ✨ Description

Calculate blending indices on the fly. Creating a blended dataset is now instantaneous rather than taking up to several minutes or even hours. Also drops some of the legacy cpp code. Part of #132.

One minor downside, there is a slight difference in the sampling order compared to the previous implementation, because Megatron-LM uses `idx*weights` instead of `(idx+1)*weights` for the target, which is arguably wrong. I tried to reproduce the old behavior for legacy datasets but couldn't get to work so I gave up. Anyway, the difference is a tiny movement of samples which won't do much, and they'll usually stay on the same batch anyway.

## 🔍 Type of change

Select all that apply:

### DIFF
--- a/fast_llm/csrc/data.cpp
+++ b/fast_llm/csrc/data.cpp
@@ -38,69 +38,6 @@ namespace py = pybind11;
 using namespace std;
 
 
-void build_blending_indices(py::array_t<int16_t>& dataset_index,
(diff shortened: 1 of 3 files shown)


tscholak (commented):
LGTM at a high level. Didn't check the exact implementation.

Issue #136, opened 2025-01-30
[feat/bug] Concatenated dataset takes too much resources


jlamypoirier:
# 🧐 Problem Description

Concatenating datasets help treating large datasets consisting of multiple files as one big dataset, but turns out it will cause us trouble, because sampling index arrays will get too big. Ex. if we're sampling 1T tokens @ 1K seqlen, avg 100 tokens/doc, we'll have:
* 10G doc idx entries (40 GB)
* 1G sample idx entries (8GB)
* 1G shuffle idx entries (4GB)

This is still somewhat manageable, but more will clearly cause memory issues. It also blocks two optimizations suggested in #132:
* Sample on GPU: may cause OOM.
* Sample in parallel: Hard to do with a single big dataset.

# 💡 Proposed Solution

A mix of two solutions:
1. Split sampling into smaller chunks (one or multiple epochs). Good for mid-sized dataset, but won't work for bigger ones if we shuffle the whole dataset. Also we'd need multiple arrays / files, so need to be careful about open file count. And only possible if we shuffle epochs independently.
2. Split the dataset into smaller chunks, i.e. go back to blending dataset chunks, sampled and shuffled independently. The most viable solution, We can still solve #25 with hierarchical blending / `concatenate_memmap`. But we have more options than before thanks to concatenated datasets and don't have to match blended chunks with actual files (ex. blend concatenations of multiple datasets, concatenate then split into chunks of fixed size independent of the file structure, etc.)

PR #146, opened 2025-02-15
Dataset from file

jlamypoirier:
# ✨ Description

Fix: #136 

Replace the both the json and concatenated memmap dataset by a generic (fast-llm) config from file. The dataset preparator now generates such config file, a blended dataset with probabilities proportional to file sizes (same as json dataset). Note:
* Dropped the json format from the preparator, but existing json datasets still work.
* Deprecated the concatenated memmap format, keeping since @oleksost is using it but we should try to remove soon.
* Added a way to split the dataset in the preparator. This was necessary to allow train/valid/test (as is done in tests)
* Updated the tutorial to use the new format. 

## 🔍 Type of change

Select all that apply:

- [x] 📝 **Documentation change** (updates documentation, including new content or typo fixes)


tscholak (commented):
Hi @jlamypoirier, thanks.
I'd like for @RaymondLi0 and @oleksost to have a look at these changes before we merge them.
Please also make sure the tests pass.

fast_llm/data/preparator/gpt_memmap/prepare.py
```diff
@@ -249,20 +251,31 @@ def run(self) -> None:
...
+            yaml.safe_dump(dataset_config, (self._config.output_path / "fast_llm_config.yaml").open("w"))
+
+            # Legacy dataset format
+            # TODO v0.3: Update docs/tutorial, then remove.
+            dataset_config = {
+                "type": "blended",
+                "datasets": [dataset_dict for dataset_dict in dataset_dicts],
+                "weights": [dataset_dict["num_tokens"] for dataset_dict in dataset_dicts],
+            }
+            yaml.safe_dump(dataset_config, (self._config.output_path / "fast_llm_config.yaml").open("w"))
```
tscholak, line 274: this is the same code as above

tscholak:
> I dropped the concatenated memmap format, assuming no one is using it yet.

In fact @oleksost is using it, so we can't just remove it.

jlamypoirier:
@tscholak @RaymondLi0 @oleksost PR is ready, can you please review?

oleksost:
Regarding deprecating `concatenated_memmap`, I had a question here: https://github.com/ServiceNow/Fast-LLM/pull/156#issuecomment-2671701857

jlamypoirier:
> Which dataset wrappers are currently available? Here we only document `blended`, `file` and `sampled`. Is the `concatenated_memmap` (will be deprecated) and `memmap` still available? Would it make sense to document each of the available wrappers?

They are still available but not expected to be used directly by normal users. `memmap` is should  to be accessed through the yaml file, and `concatenated_memmap` is to be removed once we stop using it.

> I think what's missing here is an example that shows how to turn a bunch of idx and bin files into a yaml or json dataset config file.

The yaml file is not to be created manually, it is created in dataset preparation so is already covered. Remains the case of already prepared datasets that don't have a yaml file, we can either re-prepare those or make a quick script that create it for an already prepared dataset, what do you think?

> If I have a directory with a bunch of shards (`shard_*.bin` and `shard_*.idx` files) and the corresponding `fast_llm_dataset.json` (<- generated by the prepare command) file in it, is it equivalent to use:
> 
> ```
> type: file
> path: path/to/directory/fast_llm_dataset.json
> ```
> 
> and
> 
> ```
> type: memmap
> path: path/to/directory/shard
> ```
> 
> and
> 
> ```
> type: concatenated_memmap
> path: path/to/directory/
> ```

This are all different. `file` uses the yaml file created in the `prepare` step, it won't work with older json files. Both `file` and `legacy` (json) are a probabilistic sampling of all shard.`memmap` is only a single shard (and not to be used in general), and `concatenated_memmap` is a single dataset made of all shards (and deprecated).

jlamypoirier:
@tscholak @oleksost can we merge this?

PR #148, opened 2025-02-18
Revamp feature request template for clarity and focus

tscholak:
Updates the feature Request template to ensure Fast-LLM development is efficient and team-aligned. The new template:  

- Requires clear and practical execution plans.  
- Reinforces team collaboration by explicitly asking for ticket estimates, ticket ownership, and clear criteria for acceptance.
- Uses GitHub's built-in project tracking features like labels, assignees, estimates to streamline workflow.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,33 +1,31 @@
 ---
 name: Feature request
-about: Suggest an idea for Fast-LLM
-title: "[feat] Brief description of the feature"
+about: Propose a **practical, high-impact** improvement to Fast-LLM
+title: "Brief description of the feature"
 labels: enhancement
 assignees: ''
 
 ---
 
-# 🧐 Problem Description
+# 🎯 **Goal (What & Why)**  
+> **Clearly state the purpose of this feature in one sentence.**  
+> _(Example: Add FP8 support using torchao to improve training throughput by 1.5x.)_
 
-Is your feature request related to a specific problem? Please describe it clearly.
-For example: "I'm always frustrated when [...]"
+# 🚀 **Execution Plan**  
+### **Step 1: What is the smallest working version?**  
+> _(Describe the simplest way to implement this feature with minimal effort.)_  
 
-# 💡 Proposed Solution
+### **Step 2: What additional optimizations are possible (but optional)?**  
+> _(List potential refinements that can be added in later PRs if needed.)_  
 
-Describe the solution you would like to see.
-Be as specific as possible about how it would work or be implemented.
+# 📌 **Acceptance Criteria** (Must-Haves for Completion)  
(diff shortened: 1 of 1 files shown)


jlamypoirier:
Thanks for the proposal, but I don't think it it's a good idea in its current form, because it doesn't represent a realistic workflow. Opening an issue should be about expressing a wish, we don't expect a user to know or care about implementation details and planning, which should be left to the developer actually working on the issue.

Ideally I think it should be something like:
1. A user wants a feature and opens an issue on github. At this stage we only care about results, implementation details aren't really relevant, and most details are optional.
2. We (team?) select issues to consider for development / decide whether the issue is worth looking into
3. A developer looks at the problem and comes out with a potential solution.
4. We (team?) select issues to implement / decide whether the issue is worth pursuing
5. A developer (ideally the same one) implement the issue.

The current solution is mostly about step 1 (though I still think it's too detailed), while this proposal is mostly about step 3, which I don't think is appropriate right when opening an issue. 

I agree there is value in making action plans, but I think it belongs elsewhere. Some suggestions:
* Have a mix of the existing template (simplified) and the proposed one, but clearly make the action plan as optional ("for developer use"). This would be a simple compromise.
* Have a separate template for action plans. Might be a bit confusing though.
* Add action plans later. (project board? comment on issue? edit issue?

tscholak:
Hey @jlamypoirier, thanks for the additional feedback! I think we're coming at this from different angles, so I would like to clarify a few things.

Fast-LLM is not an external, community-driven open-source project where random users submit requests. We already know the users (our team + Research + Platform AI), and we already know our strategic goals. That means feature requests are primarily internal, and we can expect a reasonable level of thought from the people proposing them (i.e. us).

I think having feature requests without scoping them properly makes execution unpredictable. If we allow issues to be too open-ended, the developer who picks it up will end up defining the scope in isolation, without visibility, input, or consultation from the team. That's why I find that a rough plan and estimate must exist before work begins. The tradeoff between effort and value must be clear to the team in order to be able to prioritize and commit to the work. Furthermore, I expect the team to be biased towards doing the simplest implementation first, so that those that depend on the feature are unblocked in the shortest amount of time. This is super important for our work.

If someone proposes a feature but isn't exactly sure how it can be implemented, that's fine. We discuss and refine it. The final implementation plan should live though in the issue, so expectations are clear for the entire team that backs it and has commit to the work being done. So, yeah, nobody is suggesting users should predefined every technical detail. But if an issue isn't scoped and lacks a clear problem definition, execution plan, and estimate, then the team can and will reject its development until it has those things.

This emphasis on team decisions is a core principle for how we will work going forward.

tscholak:
Hey @sebpaquet, I added a clarification under "Execution Plan" to reflect your point: issues can start without a full plan, but we need one before implementation. This keeps things flexible while ensuring we don't commit to open-ended work. Let me know if this works for you!

PR #177, opened 2025-03-07
optionally prevent cross-document attention

### LINKED ISSUE #62: [feat] Optional Attention Mask to Prevent Cross-Document Attention in Sequences
# 🧐 Problem Description

Currently, Fast-LLM allows self-attention across all tokens within a sequence, including tokens from different documents separated by EOS tokens. However, it was found that document-level isolation in attention can be beneficial. While models could potentially learn to ignore tokens across boundaries, this behaviour is neither guaranteed nor efficient.

# 💡 Proposed Solution

Implement an optional attention mask that prevents self-attention between different documents within a sequence.

For example, the Llama 3 paper found that such masking had limited impact during standard pre-training but proved crucial in continued pre-training on very long sequences.

# 🔄 Alternatives Considered

Just rely on the model to learn when and when not to attend across document boundaries, which is what is happening currently.

# 📈 Potential Benefits

- **Improved Performance**: Preventing unnecessary attention across unrelated documents could lead to more efficient learning during training and better inference for document-separated tasks.
- **Enhanced Flexibility**: By making this masking optional, users can tailor attention behavior to specific tasks or datasets.
- **Support for Long-Sequence Training**: As noted in the Llama 3 paper, this feature could become essential in scenarios involving very long sequences or continued pre-training.

# 📝 Additional Context

- **Reference Implementations**:
   - Packed attention mask implementation from Megatron-DeepSpeed: [GitHub Link](https://github.com/bigscience-workshop/Megatron-DeepSpeed/blob/main/megatron/utils.py#L253).
   - Flash attention updates supporting attention masks: [Dao-AILab/flash-attention#499](https://github.com/Dao-AILab/flash-attention/issues/499), [Dao-AILab/flash-attention#654](https://github.com/Dao-AILab/flash-attention/issues/654).
   - Alternatives for FlashAttention if custom masking isn't working:
       - xformers (PKU-YuanGroup/Open-Sora-Plan#109).
       - FlexAttention (https://pytorch.org/blog/flexattention/#document-maskingjagged-sequences)
- This change should be optional and integrated in a way that supports ablation studies, as there are cases (e.g., dependency-ordered document sequences) where cross-document attention may be beneficial.

sohamparikh:
# ✨ Description

Closes #62

Optionally prevent cross-document attention in packed sequences.
Introduce two boolean options in config:
1.  `data.sampling.per_document_positions`: Creates `position_ids` local to each document, instead of global `position_ids` assumed for the whole sequence. Document boundaries are determined by `tokenizer.bod_id` which requires specifying `data.tokenizer` as well
2.  `model.base_model.transformer.prevent_cross_document_attention`: Applies attention mask to each document separately using [flash_attn_varlen_func](https://github.com/Dao-AILab/flash-attention/blob/d82bbf26924c492064af8b27ab299ff4808d1bf6/flash_attn/flash_attn_interface.py#L1375).

~**Note**: This does not currently work with `sequence_data_parallel > 1`. `flash_attn_varlen_func` assumes corresponding indices in both `cu_seqlens_q` and `cu_seqlens_k` refer to the same sequence. This is not currently handled for sequence data parallelism~
Works with `sequence_data_parallel>1`

wandb runs on SmolLM-135M:
Cross-document attention (before): https://wandb.ai/akshaykalkunte/slam/runs/v6x39cllhe2xgpd1
Per-document attention (after): https://wandb.ai/akshaykalkunte/slam/runs/xpfv9w3t2gtt4z4x
image
image

## 🔍 Type of change

Select all that apply:

### DIFF
--- a/fast_llm/data/data/gpt/config.py
+++ b/fast_llm/data/data/gpt/config.py
@@ -1,13 +1,26 @@
 import logging
 
-from fast_llm.config import Field, FieldHint, check_field, config_class
+from fast_llm.config import Field, FieldHint, FieldUpdate, check_field, config_class
 from fast_llm.data.config import MultiprocessingContext, TokenizerConfig
-from fast_llm.data.data.config import DataConfig
-from fast_llm.data.dataset.gpt.config import GPTLegacyConfig, GPTLegacyDatasetConfig, GPTSampledDatasetConfig
+from fast_llm.data.data.config import DataConfig, SamplingDefaultConfig
+from fast_llm.data.dataset.gpt.config import (
+    GPTLegacyConfig,
+    GPTLegacyDatasetConfig,
+    GPTSampledDatasetConfig,
+    GPTSamplingConfig,
+    ShufflingType,
+)
 from fast_llm.engine.distributed.config import PhaseType
 from fast_llm.utils import Assert
 
 logger = logging.getLogger(__name__)
+
+
+@config_class()
+class GPTSamplingDefaultConfig(SamplingDefaultConfig, GPTSamplingConfig):
+    gpu: bool = FieldUpdate(default=True)
+    use_loss_masking_spans: bool = FieldUpdate(default=False)
+    shuffle: ShufflingType = FieldUpdate(default=ShufflingType.epoch)
 
 
 @config_class()
@@ -31,6 +44,7 @@ class GPTDataConfig(DataConfig, GPTLegacyConfig):
         desc="Configuration for the dataset(s).",
         hint=FieldHint.core,
     )
+    sampling: GPTSamplingDefaultConfig = FieldUpdate(default_factory=GPTSamplingDefaultConfig)
     data_sample_warn_time_ms: float = Field(
         default=1000,
         desc="Warn if a sample takes too long to load.",
--- a/fast_llm/data/data/gpt/config.py
+++ b/fast_llm/data/data/gpt/config.py
@@ -56,7 +56,6 @@ class GPTDataConfig(DataConfig, GPTLegacyConfig):
         desc="Multiprocessing context. Do not touch.",
         hint=FieldHint.expert,
     )
-    variable_sequence_lengths: bool | None = Field(default=None, desc="Use ")
 
     def _validate(self) -> None:
         if not self.datasets:
--- a/fast_llm/data/data/gpt/config.py
+++ b/fast_llm/data/data/gpt/config.py
@@ -47,6 +47,11 @@ class GPTDataConfig(DataConfig, GPTLegacyConfig):
         desc="Read and use loss masking spans from the dataset, if present.",
         hint=FieldHint.feature,
     )
+    per_document_positions: bool = Field(
+        default=False,
+        desc="Create separate position ids for each document in a sequence, instead of global position ids for a sequence.",
+        hint=FieldHint.feature,
+    )
 
     def _validate(self) -> None:
         if not self.datasets:
--- a/fast_llm/data/data/gpt/config.py
+++ b/fast_llm/data/data/gpt/config.py
@@ -21,6 +21,7 @@ class GPTSamplingDefaultConfig(SamplingDefaultConfig, GPTSamplingConfig):
     gpu: bool = FieldUpdate(default=True)
     use_loss_masking_spans: bool = FieldUpdate(default=False)
     shuffle: ShufflingType = FieldUpdate(default=ShufflingType.epoch)
+    per_document_positions: bool = FieldUpdate(default=False)
 
 
 @config_class()
@@ -55,16 +56,6 @@ class GPTDataConfig(DataConfig, GPTLegacyConfig):
         default=MultiprocessingContext.spawn,
         desc="Multiprocessing context. Do not touch.",
         hint=FieldHint.expert,
-    )
-    use_loss_masking_spans: bool = Field(
-        default=False,
-        desc="Read and use loss masking spans from the dataset, if present.",
-        hint=FieldHint.feature,
-    )
-    per_document_positions: bool = Field(
-        default=False,
-        desc="Create separate position ids for each document in a sequence, instead of global position ids for a sequence.",
-        hint=FieldHint.feature,
     )
 
     def _validate(self) -> None:
--- a/fast_llm/data/data/gpt/config.py
+++ b/fast_llm/data/data/gpt/config.py
@@ -21,7 +21,6 @@ class GPTSamplingDefaultConfig(SamplingDefaultConfig, GPTSamplingConfig):
     gpu: bool = FieldUpdate(default=True)
     use_loss_masking_spans: bool = FieldUpdate(default=False)
     shuffle: ShufflingType = FieldUpdate(default=ShufflingType.epoch)
-    per_document_positions: bool = FieldUpdate(default=False)
 
 
 @config_class()
@@ -57,6 +56,7 @@ class GPTDataConfig(DataConfig, GPTLegacyConfig):
         desc="Multiprocessing context. Do not touch.",
         hint=FieldHint.expert,
     )
+    variable_sequence_lengths: bool | None = Field(default=None, desc="Use ")
 
     def _validate(self) -> None:
         if not self.datasets:
--- a/fast_llm/data/data/gpt/data.py
+++ b/fast_llm/data/data/gpt/data.py
@@ -35,14 +35,14 @@ class GPTBatch:
 
 
 def gpt_data_collate_fn(
-    batch: list[GPTSample], use_loss_masking_spans: bool, use_document_boundaries: bool
+    batch: list[GPTSample], use_loss_masking_spans: bool, cross_document_attention: bool
 ) -> GPTBatch:
     stacked_ids = np.stack([sample.token_ids for sample in batch])
     stacked_spans = None
     sequence_lengths = None
     if use_loss_masking_spans:
         stacked_spans = [torch.from_numpy(sample.loss_masking_spans) for sample in batch]
-    if use_document_boundaries:
+    if not cross_document_attention:
         sequence_lengths = [torch.tensor(sample.sequence_lengths) for sample in batch]
     return GPTBatch(
         token_ids=torch.from_numpy(stacked_ids), loss_masking_spans=stacked_spans, sequence_lengths=sequence_lengths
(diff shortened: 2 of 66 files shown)


jlamypoirier (changes requested):

fast_llm/models/gpt/model.py
```diff
@@ -228,7 +234,26 @@ def preprocess(
             else:
                 # TODO: Avoid multiple contiguous calls?
                 tokens = batch.token_ids[:, sequence_k - sequence_q : sequence_k].contiguous()
-
+            if batch.position_ids is not None:
```
jlamypoirier, line 237: This is only active with `sequence_first`, so you should ensure `sequence_first` is `True` for varlen . See `preprocessed_meta`

fast_llm/data/dataset/gpt/sampled.py
```diff
@@ -491,7 +498,14 @@ def __getitem__(self, idx: int) -> typing.Any:
             spans = np.stack(spans, dtype=np.int32)
         else:
             spans = None
-        return GPTSample(token_ids=token_ids, loss_masking_spans=spans)
+        if self._config.per_document_positions:
+            start_indices = np.where(token_ids == self._tokenizer.bod_id)[0][:-1]
```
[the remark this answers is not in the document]
sohamparikh: I changed it to using the bos token since we wanted to also make it work with sequences packed and padded outside of fast-llm, as a quick and dirty hack. Although I think that's not a priority anymore, since I'm also working on a padding PR this week.

@tscholak do you think there is still a reason to do this or can we just wait for padding to be available in fast-llm as well?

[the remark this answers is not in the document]
tscholak: Thanks @sohamparikh, let's wait for padding :)

fast_llm/layers/language_model/embedding.py
```diff
@@ -76,7 +75,6 @@ def __init__(
 
     @torch.compile
     def _forward(self, input_: torch.Tensor, position_ids: torch.Tensor | None) -> torch.Tensor:
-        Assert.eq(position_ids is not None, self._use_absolute_position_embeddings)
```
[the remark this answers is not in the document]
jlamypoirier: Preferably yes, this is for extra safety

fast_llm/models/gpt/model.py
sohamparikh: I'm confused, this is executed regardless of `sequence_first` currently. varlen shouldn't depend on it either

jlamypoirier: Ah I misread the diff, nevermind

fast_llm/data/dataset/gpt/sampled.py
```diff
@@ -337,9 +338,14 @@ def __getitem__(self, index: int) -> typing.Any:
         loss_masking_spans = (
             np.stack(loss_masking_spans, dtype=np.int32) if self._config.use_loss_masking_spans else None
         )
+        position_ids = (
```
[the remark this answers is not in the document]
sohamparikh: i did it since `position_ids` are more commonly understood. Makes sense to use seqlens instead

sohamparikh:
@jlamypoirier I've addressed your comments:
1. Simpler config: `batch.variable_sequence_lengths` is the only option
2. Use `seqlens` instead of `position_ids` in `GPTSample` and `GPTBatch`. `position_ids` are only created for absolute position embeddings
3. `cu_seqlens_*` and `max_seqlens_*` are now created inside `FlashAttnVarlenPreprocessor`. This is to mainly avoid `preprocess` getting too big and messy
4. Removed flash_attn wrapper
5. Hacked a workaround for `sequence_data_parallel > 1` with varlen (see loss curves below). It's not super efficient, but should work okay for lower parallelism


image

fast_llm/layers/transformer/attention.py
```diff
@@ -340,16 +343,57 @@ def forward(self, input_: torch.Tensor, kwargs: dict[str, typing.Any]) -> tuple[
...
-                causal=True,
-                generator=self._tensor_space.distributed.tp_generator,
-                softmax_scale=self._softmax_scale,
-            ).flatten(-2)
+            assert _flash_available
+            with set_generator(self._tensor_space.distributed.tp_generator):
+                if (cu_seqlens_q := kwargs.get(TransformerKwargs.cu_seqlens_q, None)) is not None:
+                    out_dims = query.size()
+                    query = query.view(-1, query.size(-2), query.size(-1))
+                    if key.size(-3) != out_dims[-3]:
```
jlamypoirier, line 351: I'm not following, what is this for exaclty?

fast_llm/layers/transformer/preprocessing.py
```diff
@@ -254,3 +260,73 @@ def preprocess_meta(self, kwargs: dict[str, typing.Any]) -> None:
...
+
+
+class FlashAttnVarlenPreprocessor:
+    def __init__(self, config: TransformerConfig, tensor_space: TensorSpace):
+        self._config = config
+        self._tensor_space = tensor_space
+        self._distributed_config = self._tensor_space.distributed_config
+        assert self._config.do_use_flash_attention(self._distributed_config)
+
+    def preprocess(self, kwargs: dict[str, typing.Any]) -> None:
```
jlamypoirier, line 272: That's a big one, consider adding comments and maybe a test or two?

jlamypoirier (commented):
Looks good, got a few comments on the details

fast_llm/data/data/gpt/config.py
```diff
@@ -56,11 +56,7 @@ class GPTDataConfig(DataConfig, GPTLegacyConfig):
         desc="Multiprocessing context. Do not touch.",
         hint=FieldHint.expert,
     )
-    use_loss_masking_spans: bool = Field(
-        default=False,
-        desc="Read and use loss masking spans from the dataset, if present.",
-        hint=FieldHint.feature,
-    )
+    variable_sequence_lengths: bool | None = Field(default=None, desc="Use ")
```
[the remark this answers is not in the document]
sohamparikh: this is unused, `GPTSamplingData.variable_sequence_lengths` is used in `GPTSampledIndexedDataset` and `LegacyGPTSampledIndexedDataset`

fast_llm/layers/transformer/attention.py
sohamparikh: it's a check for `sequence_data_parallelism`. If number of key/query tokens are the same, we want to avoid the extra computation in this block

fast_llm/data/dataset/gpt/sampled.py
```diff
@@ -491,7 +500,15 @@ def __getitem__(self, idx: int) -> typing.Any:
             spans = np.stack(spans, dtype=np.int32)
         else:
             spans = None
-        return GPTSample(token_ids=token_ids, loss_masking_spans=spans)
+        seqlens = (
+            np.array(
+                [sample.token_ids.size - (idx == len(sample_list) - 1) for idx, sample in enumerate(sample_list)],
```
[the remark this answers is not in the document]
jlamypoirier: I see, thanks!

fast_llm/layers/transformer/attention.py
sohamparikh: For some more context, the varlen function expects the same number of sequences for key/queries, so this block slices kv pairs accordingly (instead of including all past kv pairs)

fast_llm/layers/transformer/preprocessing.py
sohamparikh: done

fast_llm/data/dataset/gpt/config.py
```diff
@@ -62,6 +62,11 @@ class GPTSamplingConfig(SamplingConfig):
         desc="Shuffling strategy.",
         hint=FieldHint.feature,
     )
+    per_document_positions: bool | None = Field(
```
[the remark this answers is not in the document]
sohamparikh: Now that we have a single flag, does it make sense to revert to something more indicative like `prevent_cross_document_attention`, or `document_local_attention`? Because we're still packing variable length sequences even with `variable_sequence_length=False`

@tscholak do you have any recommendations?

jlamypoirier (changes requested):
~I don't think your approach for sequence-data-parallel makes sense. It simply splits into independent sequences, so it's just plain data parallelism. It's also very inefficient, and only works when there are lots of small sequences, in which case sequence-data-parallel is useless to begin with.~

~What you need to do instead is adjust `cu_seqlens_k` to the correct values, `_flash_attn_varlen_func` is made for this and will handle the rest~

Edit: I think I got your implementation wrong, looking again

fast_llm/layers/transformer/attention.py
jlamypoirier: Can't you just set `seqlen_q=0` for the sequences outside the current range?

sohamparikh: yes, that's much simpler. Thanks!

fast_llm/data/data/gpt/data.py
```diff
@@ -58,6 +66,7 @@ def __init__(
         distributed_config: DistributedConfig,
         vocab_size: int,
         max_sequence_length: int,
+        use_document_boundaries: bool | None = None,
```
jlamypoirier, line 69: Not convinced about this new name. How about `separate_documents`? Or `separate_sequences`, to make it non-specific to language models?

fast_llm/layers/transformer/config.py
```diff
@@ -473,6 +478,11 @@ class TransformerConfig(TransformerArchitectureConfig, BaseModelConfig):
     use_flash_attention: bool = Field(
         default=True, desc="Enable Flash Attention if possible.", hint=FieldHint.optional
     )
+    isolate_document_context: bool = Field(
```
jlamypoirier, line 481: I think it was better in the BatchConfig, since it's only used for preprocessing and not by the transformer directly. Also same comment about naming.

jlamypoirier (approved):
Some minor comments, otherwise LGTM. Great work!

fast_llm/layers/transformer/config.py
tscholak: Let's call this `enable_cross_document_attention`, flip the boolean and make the default `True`.

jlamypoirier: Guess that works. Or simply `cross_document_attention`

PR #179, opened 2025-03-07
Multi-token prediction

### LINKED ISSUE #167: Multi-token prediction
# 🎯 **Goal (What & Why)**
**Enable multi-token prediction in Fast-LLM with minimal memory overhead.**

Instead of predicting just the next token, the GPT model will be modified to predict multiple future tokens in parallel using independent output heads. The number of language modelling heads will be configurable via `transformer.lm_heads`, and the total loss will be the sum of the individual prediction losses.

Multi-token prediction has been shown to significantly improve performance on generative tasks like coding and algorithmic reasoning, as highlighted in [Better & Faster Large Language Models via Multi-token Prediction](https://arxiv.org/abs/2404.19737). Key benefits include:  
- Better downstream performance: Multi-token models consistently outperform strong baselines, especially in tasks requiring multi-step reasoning (e.g., code generation).
- Scalability to large models: Gains are most pronounced for models 1B+ in size.
- No additional training cost: Training time remains unchanged, provided memory overhead is managed properly.
- Inference speedup with self-speculative decoding:
    - Up to 3 times faster inference for code.
    - Up to 2.7 times speedup for text.

This feature allows experimentation with multi-token prediction while keeping memory usage low and training speed unchanged.

# 🚀 **Execution Plan**

### **Step 1: Minimal Working Version**
- Introduce a configuration option `transformer.lm_heads` to specify how many future tokens should be predicted (default: 1).
- Modify the model to support multiple independent output heads operating on a shared transformer trunk.
- Ensure that loss is computed as the sum of the `transformer.lm_heads` next-token prediction losses.
- Optimize memory usage by ensuring that:
  - The shared GPT trunk is computed once and reused across all heads.
  - Each output head performs its forward and backward pass sequentially, without keeping unnecessary intermediate activations in memory.
  - Only the final accumulated gradient from all heads is applied to the shared transformer trunk.

* The feature must be **functional** and tested with minimal memory overhead.
* The implementation must ensure **no regression in training speed**.
* The implementation must be **documented clearly** to guide future experiments with multi-token prediction.

# 🛠️ **Project Management**

RaymondLi0:
# ✨ Description
Add a minimally working version of Multi-token prediction:
- a new `num_multi_token_prediction_heads` argument. `None` by default. If true, will add new layers for multi-token prediction, that replace the standard `lm_head` layer. For each token to predict, we add a `MultiTokenPredictionTransformerLayer` and a `MultiTokenPredictionLanguageModelHead`
- All the output weights are shared. Like before, they can be shared with the input embeddings with `tie_word_embeddings`
- Each multi-token prediction loss is tracked separately, and contributes equally to the final loss being optimized.
- At export the additional lm-heads are discarded, resulting in a transformer model with `num_layers + 1` layers.

In a future PR:
- Add more labels in data-sampling to avoid input-truncation
- add weights for each future-token prediction loss (e.g. the next token could be more important than the future ones)
- handle sequence-parallelism and cross-entropy splits
- Convert the additional lm-heads (requires a corresponding hf-transformer implementation)
- Improve throughput

Closes #167 

## 🔍 Type of change

Select all that apply:

## Sanity check
A baseline transformer model reaches the same loss as a model with `num_multi_token_prediction_heads=1` and one less layer in the shared trunk. Memory usage is very similar.
Screenshot 2025-03-24 at 4 51 29 PM
Screenshot 2025-03-24 at 4 57 46 PM

## Performance
The additional layers seem to negatively impact training throughput
Screenshot 2025-03-24 at 4 53 31 PM

---

## 🗒️ Additional Notes


fast_llm/layers/language_model/multi_token_prediction_head.py
```diff
@@ -0,0 +1,132 @@
...
+        # TODO: Torch compile implementation sometimes break.
+        # TODO: Double-check correctness, optimize a bit more.
+        # TODO: Drop autograd entirely.
+        # TODO: Skip cross-entropy backward if not needed.
+        language_model_loss = self._forward(transformer_layer_output, kwargs, losses)
+        if language_model_loss is not None:
+            losses[self.loss_name].append(language_model_loss)
+        if self.is_last_head:
+            # Last layer should return the loss for backward.
+            return language_model_loss
```
jlamypoirier, line 89: This returns the loss for the most distant predicted token, which doesn't make much sense. How about running the predictions in reversed order so the last one is the next token (index=0)? This would make the return value here more relevant and merge the `is_last_head` and `multi_token_prediction_index > 0` conditions.

fast_llm/layers/language_model/multi_token_prediction_head.py
```diff
@@ -0,0 +1,132 @@
...
+        metrics: dict[str, typing.Any] | None = None,
+    ) -> torch.Tensor:
+        # TODO MTP: fix dims?
+        if isinstance(input_, TensorMeta):
+            return self._get_meta(input_, "output", kwargs)
+        output = super().forward(input_, kwargs, losses, metrics)
+        return torch.stack((input_, output), dim=0)
+
+
+class MultiTokenPredictionLanguageModelHead(LanguageModelHead):
```
jlamypoirier, line 40: Is this really worth a separate class? Seems like we could just add a `multi_token_prediction_index`/`_prediction_distance` option to the base class, it would only lead to tiny changes.

fast_llm/models/gpt/model.py
```diff
@@ -71,6 +75,30 @@ def __init__(
         else:
             self._flash_varlen_preprocessor = FlashAttnVarlenPreprocessor(self._config.transformer, self._tensor_space)
 
+    def get_language_model_layers(self) -> list[Layer]:
+        if self._config.num_multi_token_prediction_heads:
+            return [
+                layer
+                for i in range(self._config.num_multi_token_prediction_heads)
+                for layer in [
+                    MultiTokenPredictionTransformerLayer(
```
jlamypoirier, line 84: The last one can be a plain `TransformerLayer`, no need for stacking. And if the next token is predicted last, it can be followed by a plain `LanguageModelHead`

fast_llm/layers/language_model/config.py
```diff
@@ -57,6 +61,12 @@ class LanguageModelArchitectureConfig(BaseModelArchitectureConfig):
     tie_word_embeddings: bool = Field(
         default=True, desc="Tie the output weights (logits) with the vocabulary embedding.", hint=FieldHint.core
     )
+    num_multi_token_prediction_heads: int | None = Field(
+        default=None,
+        desc="Number of multi-token prediction heads.",
+        hint=FieldHint.feature,
+        valid=skip_valid_if_none(check_field(Assert.gt, 0)),
```
jlamypoirier, line 68: 1 is a weird case where we just add an extra transformer layer to a standard LM for some reason, how about `skip_valid_if_none(check_field(Assert.gt, 1))`?

jlamypoirier (commented):
Looks good, but I have some suggestions to make it behave more like a standard language model with added prediction heads rather than an entirely new thing.

fast_llm/data/dataset/gpt/sampled.py
```diff
@@ -128,6 +128,7 @@ def _sample(self) -> None:
         # Calculate basic stats.
         documents_per_epoch = document_sizes.numel()
         tokens_per_epoch = document_sizes.sum().item()
+        # TODO MTP: Produce more labels to provide labels for the multi-token prediction heads?
```
tscholak, line 131: I'm not following, could you explain what you mean here?
Surely you're shifting the sequence for the additional heads, are you not?

fast_llm/layers/language_model/multi_token_prediction_head.py
tscholak: Screenshot 2025-03-24 at 9 37 30 PM

we want this ^^^ because this way we ensure that memory impact is minimal.

tscholak (commented):
Looks good, thanks @RaymondLi0!
I wonder though if it wouldn't be better to avoid adding the additional transformer layer for the first lm head. If that was the case, then the mtp feature could always be on (with default 1 head) and would do the same thing as before. Only for more than 1 head we would see additional transformer layers being added.

fast_llm/layers/language_model/multi_token_prediction_head.py
jlamypoirier: We do want to compute the heads one at the time, but the order in which we do it doesn't matter.

RaymondLi0: For the conditions: the output-weights should be defined in the first layer (according to https://github.com/ServiceNow/Fast-LLM/blob/f7c9e65e7bdaeb5c383746b95c7cdaea06446769/fast_llm/engine/base_model/base_model.py#L129), right? And the last layer should return its loss. So even if we do the reverse order I don't think we can merge the `is_last_head` and `multi_token_prediction_index > 0` conditions.

I also thought it would make it convenient if later we want to support a sequential version of this, where each head would use the output of the previous one as input (this is what deepseek-v3 does for example)

fast_llm/layers/language_model/head.py
```diff
@@ -151,6 +160,7 @@ def _logits_cross_entropy_forward_backward_split(
                 return None, None
         else:
             loss = None
+            # TODO MTP: allow a _cross_entropy_splits that is not a divisor of the sequence length
```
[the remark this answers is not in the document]
RaymondLi0: Because for the additional heads, the input is truncated, so the sequence-length at that point will usually not be a multiple of `_cross_entropy_splits`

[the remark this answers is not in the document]
RaymondLi0: This can also be addressed by creating more labels at the data-sampling phase to avoid truncation of the sequence

fast_llm/layers/language_model/multi_token_prediction_head.py
RaymondLi0: Hmm, we could yes, but there are quite a bit of changes, especially in `_foward_backward`, with the shifting/truncating of labels and inputs, and the potential handling of sequence-parallel and cross-entropy-splits (not supported now, but this would need different code).
Do you still think we should put this in the base class `LanguageModelHead` ?

RaymondLi0: Oh I guess the logic of the base class would just be a special-case where `multi_token_prediction_index=0`
Then indeed I could just add this code to the base class

RaymondLi0: I could also remove the `MultiTokenPredictionTransformerLayer` class by adding a `stacked_output` parameter to the `TransformerLayer` class. wdyt @jlamypoirier @tscholak ?

fast_llm/layers/language_model/config.py
RaymondLi0: `num_multi_token_prediction_heads=1` is the default now, to behave like a standard LM. So keeping this condition, and removed the `skip_valid_if_none`

fast_llm/layers/language_model/multi_token_prediction_head.py
RaymondLi0: I removed both `MultiTokenPredictionTransformerLayer` and `MultiTokenPredictionLanguageModelHead` and moved the code to the base classes. Things should be a bit simpler now.

fast_llm/layers/language_model/config.py
```diff
@@ -22,6 +22,10 @@ class LanguageModelLossNames:
     language_model_loss = "language_model_loss"
     z_loss = "z_loss"
 
+    @classmethod
```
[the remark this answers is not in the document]
RaymondLi0: Would make more sense as a `@staticmethod`?

fast_llm/layers/language_model/config.py
```diff
@@ -57,6 +61,12 @@ class LanguageModelArchitectureConfig(BaseModelArchitectureConfig):
     tie_word_embeddings: bool = Field(
         default=True, desc="Tie the output weights (logits) with the vocabulary embedding.", hint=FieldHint.core
     )
+    num_multi_token_prediction_heads: int | None = Field(
```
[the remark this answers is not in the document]
RaymondLi0: @tscholak  educated me about our naming conventions https://servicenow.github.io/Fast-LLM/contributing/style-guide/#naming-conventions let's go with `prediction_heads` then

fast_llm/layers/language_model/multi_token_prediction_head.py
jlamypoirier: I guess you're right. I'll let you decide on the best option here.

fast_llm/models/gpt/model.py
```diff
@@ -71,6 +71,28 @@ def __init__(
         else:
             self._flash_varlen_preprocessor = FlashAttnVarlenPreprocessor(self._config.transformer, self._tensor_space)
 
+    def get_output_layers(self) -> list[Layer]:
+        return [
+            layer
+            for i in range(self._config.prediction_heads)
+            for layer in [
+                TransformerLayer(
```
jlamypoirier, line 79: Wrong when num_layers=0. (Mostly for debug but we do want to support it.)

fast_llm/models/gpt/conversion.py
```diff
@@ -160,58 +161,107 @@ def _create_config_converters(cls) -> list[ParamConverter]:
...
+        converters += self._get_weight_and_bias_converters(
+            f"layers.{num_layers + 1}.final_norm", "model.norm", norm_bias
+        )
+        # Output weights
+        if self._model.config.base_model.tie_word_embeddings:
+            converters.append(IgnoreImportWeightConverter((), "lm_head.weight"))
+        else:
+            converters.append(WeightConverter(f"layers.{num_layers + 1}.output_weights", "lm_head.weight"))
+
+        # MTP-heads > 0 are thrown away
```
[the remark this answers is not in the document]
tscholak: Eventually, we want to export the MTP layers as well if the export target supports that architecture. (This is out of scope for this PR.) Otherwise, the heads are discarded, and a warning should be emitted.

fast_llm/models/gpt/model.py
tscholak: the whole MTP thing doesn't make sense for num_layers=0. we would have 0 transformer layers for the first head and each 1 for every additional head.

tscholak (approved):
Thanks @RaymondLi0, this looks great.
I see some logic in trying to avoid truncation of the inputs when additional heads are present, but I would time box this at this point. I suggest 2 hours max. I'd like us to quickly move on to running actual experiments that validate the value proposition of MTP before over-investing in implementation details. Thanks.

fast_llm/engine/checkpoint/state_dict.py
```diff
@@ -56,7 +56,7 @@ def save(self, config: CheckpointSaveConfig, metadata: CheckpointMetadata) -> No
                 saver.add_tensor(self._get_key(exported_name, shard_name), exported_tensor)
 
         for shard_name, shard_state_dict in state_dict.items():
-            assert not shard_state_dict, (shard_name, list(state_dict))
+            assert not shard_state_dict, f"Un-handled entries after conversion: {(shard_name, list(shard_state_dict))}"
```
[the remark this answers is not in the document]
RaymondLi0: Ok, I think I changed this because previously this printed something like `('weights', ['weights'])`
I guess we could just print `state_dict` instead?

fast_llm/layers/language_model/config.py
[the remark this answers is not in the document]
RaymondLi0: It's for the same purpose that we have `language_model_loss = "language_model_loss"` above, no?

fast_llm/layers/language_model/head.py
```diff
@@ -56,23 +59,24 @@ def __init__(
 
         hidden_dim = self._tensor_space.get_tensor_dim(TransformerDimNames.hidden)
 
+        self._loss_name = LanguageModelLossNames.multi_token_prediction_loss(prediction_distance)
```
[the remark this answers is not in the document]
RaymondLi0: Renamed to `language_model_loss`

fast_llm/data/dataset/gpt/sampled.py
RaymondLi0: Ok I can have a quick look at this this afternoon, and let's try to merge this PR afterwards

jlamypoirier (approved):
Minor suggestion, otherwise LGTM

fast_llm/models/gpt/model.py
```diff
@@ -71,7 +71,35 @@ def __init__(
...
+                    self._config,
+                    self._tensor_space,
+                    prediction_distance=i,
+                ),
+            ]
+        ]
+
     def get_layers(self) -> list[Layer]:
+        if self._config.transformer.num_layers == 0:
+            Assert.eq(self._config.prediction_heads, 1)
```
jlamypoirier, line 98: This could be checked in config validation

tscholak: for a debug case like num_layers=0 that has no practical applications?
I think not.

fast_llm/data/dataset/gpt/sampled.py
RaymondLi0: Will merge now, and add this to a new PR, since I'd need to test it a little bit more

RaymondLi0:
Thanks for the reviews @jlamypoirier @tscholak ! I'll move the additional labels to a new PR to respect the time-boxing.

PR #187, opened 2025-03-14
Fix bugs in cross-attention

sohamparikh:
# ✨ Description

Fix bugs from the last commit in https://github.com/ServiceNow/Fast-LLM/pull/177. Additionally fix for tests.
Oversight on my part. Verified that the failing tests pass and training runs without errors


jlamypoirier (approved):
Thanks for the quick fix

PR #188, opened 2025-03-14
[feat] Track entropy and MI of routing distribution for topk MoE

oleksost:
# ✨ Description
To better detect potential routing collapse and have a better understanding about the routing distribution, we can track the average entropy and mutual information (which is the difference between the entropy of average routing and average routing entropy) of routing probabilities.

Collapse routing would have low entropy and low mutual information. A healthy and specialised router would have low entropy and high mutual information. 

## 🔍 Type of change

Select all that apply:

## 📝 Changes
- added calculation of both metrics in the `mixture_of_experts.py`, they are calculated only for the topk routing type. 

### General

- [x] 📜 I have read and followed the [contributing guidelines](https://servicenow.github.io/Fast-LLM/developers/contributing).
- [x] 🏷️ I am using a clear and descriptive PR title that summarizes the key change or feature introduced.
- [x] 🎉 The functionality is complete, and I have tested the changes.
- [x] ⚠️ The change does not introduce any new issues (e.g., runtime warnings, type checker errors, linting problems, unhandled edge cases).
- [x] 🧩 I have commented my code, especially in hard-to-understand areas.

### Testing

- [x] 🧪 I have added or updated tests to cover my changes.
- [x] 🚦 I have tested these changes on GPUs and verified training stability.
- [x] 🏋️ I have tested the changes on realistic training workloads, if applicable.

### Performance Impact

I am not 100% sure there is no performance impact, we are calculating the stats at each forward pass through the router.

---

## 🗒️ Additional Notes

### DIFF
--- a/fast_llm/layers/transformer/mixture_of_experts.py
+++ b/fast_llm/layers/transformer/mixture_of_experts.py
@@ -25,30 +25,36 @@ from fast_llm.utils import Assert
 
 logger = logging.getLogger(__name__)
 
+
 def calculate_normalized_average_entropy(probs: torch.Tensor) -> torch.Tensor:
-    '''
+    """
     Calculates routing entropy for each token, then averages over all tokens.
     If low, means a lot of mass is put on a single expert, which can indicate collapse.
-    '''
+    """
     n_experts = probs.size(-1)
     entropy_values = entropy(probs)
     average_entropy = entropy_values.mean()  # Average over batch and tokens
     return average_entropy / torch.log(torch.tensor(n_experts, dtype=probs.dtype))
 
+
 def entropy(probs: torch.Tensor) -> torch.Tensor:
     probs = torch.clamp(probs, min=1e-9)  # Avoid log(0)
     return -torch.sum(probs * torch.log(probs), dim=-1)
 
+
 def calculate_mutual_information(probs: torch.Tensor) -> torch.Tensor:
-    '''
-    Calculates the difference between the entropy of the average routing and 
+    """
+    Calculates the difference between the entropy of the average routing and
     the average routing entropy. If low, means that routing is not informative.
-    '''
+    """
     average_routing = torch.mean(probs, dim=1)  # Average over tokens
     entropy_avg_routing = entropy(average_routing).mean()  # H[E[X]], mean over batch
     entropy_routing = entropy(probs).mean()  # E[H[X]]
-    
-    return (entropy_avg_routing - entropy_routing) / torch.log(torch.tensor(probs.size(-1), dtype=probs.dtype))  # Normalize
+
+    return (entropy_avg_routing - entropy_routing) / torch.log(
+        torch.tensor(probs.size(-1), dtype=probs.dtype)
+    )  # Normalize
+
 
 class MixtureOfExpertMLP(MLPBase):
     """
@@ -135,7 +141,6 @@ class MixtureOfExpertMLP(MLPBase):
         else:
             raise NotImplementedError(self._routing_type)
 
-
         if self._debug_mode:
             # To log all ranks set `global_=False`
             self._debug_log(scores, "Router scores", TransformerDimNames.top_experts, kwargs)
@@ -203,13 +208,13 @@ class MixtureOfExpertMLP(MLPBase):
             # Calculate and log entropy and mutual information
             entropy = calculate_normalized_average_entropy(probs)
             mutual_info = calculate_mutual_information(probs)
-            
+
             # Store these metrics
             if "router_entropy" not in losses:
                 losses["router_entropy"] = []
             if "router_mutual_info" not in losses:
                 losses["router_mutual_info"] = []
-                
+
             losses["router_entropy"].append(entropy.detach())
             losses["router_mutual_info"].append(mutual_info.detach())
 
--- a/fast_llm/layers/transformer/mixture_of_experts.py
+++ b/fast_llm/layers/transformer/mixture_of_experts.py
@@ -29,7 +29,7 @@ logger = logging.getLogger(__name__)
 def calculate_normalized_average_entropy(probs: torch.Tensor) -> torch.Tensor:
     """
     Calculates routing entropy for each token, then averages over all tokens.
-    If low, means a lot of mass is put on a single expert, which can indicate collapse.
+    If low, means a lot of mass is put on a single expert in all tokens, which can indicate collapse or specialization.
     """
     n_experts = probs.size(-1)
     entropy_values = entropy(probs)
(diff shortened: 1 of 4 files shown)


tscholak (commented):
idea is good, thanks @oleksost.
bit weird that all these metrics are appearing as losses. that name should be reserved for things for which gradients are computed. just call this dict metrics?

oleksost:
Yes, indeed, this has to be improved.

fast_llm/engine/base_model/base_model.py
```diff
@@ -135,3 +135,7 @@ def get_tied_weights(self) -> dict[str, tuple[ParameterMeta, tuple[int, ...]]]:
     @abc.abstractmethod
     def loss_defs(self) -> list[LossDef]:
         pass
+
+    @property
```
jlamypoirier, line 139: This loss/metric split is way more complicated than needed. How about having a single entry, and using a `is_metric` flag in `LossDef` (or a derived class) to distinguish? Then no change is needed other than extracting metrics from the context before returning from `run_step`

fast_llm/engine/schedule/runner.py
```diff
@@ -289,6 +312,19 @@ def _reduce_losses(self, context: BatchContext) -> dict[str, float | int]:
             for name, reduced_loss in reduced_losses.items()
         }
 
+    def _is_reduced_metric(self, metric_name: str) -> bool:
+        """Check if a metric should be reduced (is defined in a TransformerReducedMetrics subclass)."""
+        from fast_llm.layers.transformer.config import TransformerReducedMetrics
```
jlamypoirier, line 317: We can't use hard-coded values here. Suggestion above would fix it, or there are a few other ways to get this dynamically.

fast_llm/layers/transformer/mixture_of_experts.py
```diff
@@ -169,11 +199,27 @@ def _topk_routing(
...
+        metrics: dict | None = None,
     ) -> tuple[torch.Tensor, torch.Tensor]:
         top_logits, top_experts = torch.topk(logits, k=self._experts_per_token, dim=-1)
         scores = torch.softmax(top_logits, dim=-1, dtype=torch.float32)
         if losses is not None or (self.training and grad_scale is not None):
             probs = torch.softmax(logits, dim=-1, dtype=torch.float32)
+
+
+            # Store these metrics        
+            if metrics is not None:
```
jlamypoirier, line 211: Given the extra computation involved, this should be enabled through a config parameter

tests/test_moe_metrics.py
```diff
@@ -0,0 +1,282 @@
...
+                [0.1, 0.7, 0.1, 0.1],  # Token 3: mostly expert 1
+            ],
+        ]
+    )
+    mutual_info = calculate_mutual_information(mixed_probs)
+    # The expected value is between 0 and 1
+    assert 0.0 < mutual_info < 1.0, f"Expected value between 0 and 1, got {mutual_info}"
+
+
+def test_edge_cases():
```
jlamypoirier, line 157: More explicit name?

tests/test_moe_metrics.py
```diff
@@ -0,0 +1,282 @@
...
+    small_probs[:, :, 0] = 1.0 - 3e-8  # Make sure they sum to 1
+    norm_entropy = calculate_normalized_average_entropy(small_probs)
+    mutual_info = calculate_mutual_information(small_probs)
+    assert torch.isclose(norm_entropy, torch.tensor(0.0), atol=1e-5), f"Expected ~0.0, got {norm_entropy}"
+    assert torch.isclose(mutual_info, torch.tensor(0.0), atol=1e-5), f"Expected ~0.0, got {mutual_info}"
+
+
+
+@pytest.fixture
+def setup_runner():
```
jlamypoirier, line 177: These don't belong here. How about `test_runner.py`?

tests/test_moe_metrics.py
```diff
@@ -0,0 +1,282 @@
...
+    
+    reduced_losses = runner._reduce_losses(context)
+    
+    assert "test_loss" in reduced_losses
+    assert pytest.approx(reduced_losses["test_loss"], 0.001) == 2.0
+
+
+
+
+if __name__ == "__main__":
```
jlamypoirier, line 281: Not needed

fast_llm/layers/transformer/mixture_of_experts.py
```diff
@@ -26,6 +27,35 @@
 logger = logging.getLogger(__name__)
 
 
+def calculate_normalized_average_entropy(probs: torch.Tensor) -> torch.Tensor:
```
jlamypoirier, line 30: Could try `@torch.compile` on these for a free performance boost.

fast_llm/layers/transformer/mixture_of_experts.py
```diff
@@ -26,6 +27,35 @@
...
+    """
+    Calculates routing entropy for each token, then averages over all tokens.
+    If low, means a lot of mass is put on a single expert in all tokens, which can indicate collapse or specialization.
+    """
+    n_experts = probs.size(-1)
+    entropy_values = entropy(probs)
+    average_entropy = entropy_values.mean()  # Average over batch and tokens
+    return average_entropy / torch.log(torch.tensor(n_experts, dtype=probs.dtype, device=probs.device))
+
+def entropy(probs: torch.Tensor) -> torch.Tensor:
```
jlamypoirier, line 40: `calculate_entropy`

jlamypoirier (commented):
Looks good, got some comments on the structure.

tests/test_moe_metrics.py
tscholak: why don't they belong here? this is fixture is only useful for the tests in this suite.

fast_llm/layers/transformer/mixture_of_experts.py
tscholak: how much compute are we talking about for these metrics? likely this won't be noticeable.

jlamypoirier:
@oleksost Are you still working on this?

oleksost:
@jlamypoirier yes, will address your comments today. Sorry, it was prioritised in favour of mamba.

tests/test_moe_metrics.py
oleksost: Maybe we can move it to common.py in the future, it maybe be reused by other tests (e.g. ssm test)

fast_llm/layers/transformer/mixture_of_experts.py
oleksost: this is already controlled by `training.logs.interval` parameter afaiu. Do you think we need a seperate parameter for MoE stats logging?

fast_llm/engine/schedule/runner.py
oleksost: Simplified the setup s.t. all metrics that come back from a forward pass are reduced automatically, hence no need for this function.

fast_llm/engine/base_model/base_model.py
oleksost: This would be nice! 

Maybe better to leave it for a separate pr? It would make this one larger as it would require also changing the interfaces of the models' forward functions (that expect losses and metrics) as well as making sure that metrics are only calculated when `return_metrics` is True.

tests/test_moe_metrics.py
jlamypoirier: I'm talking about the associated tests, not the fixture. They test a feature of `ScheduleRunner` and have nothing to do with MoE other than the loss names which aren't relevant to the tests.

fast_llm/layers/transformer/mixture_of_experts.py
jlamypoirier: Ideally we'd like to calculate the bare minimum by default (and the computation isn't optimized), so I think yes.

If we really want by default I guess we could have a parameter with that defaults to true but can disabled for performance, eg. for benchmarks.

oleksost: addressed

oleksost:
@jlamypoirier I think I addressed all the comments.

fast_llm/engine/base_model/base_model.py
jlamypoirier: There isn't much change needed actually, just need to add `kwargs["return_metrics"]`. I would prefer doing this here so we don't grow `ScheduleRunner` too much.

Issue #212, opened 2025-03-27
Frozen reference model support for DPO, distillation, etc.


tscholak:
# 🎯 **Goal (What & Why)**

Add support for loading and executing a (potentially very large) frozen reference model in parallel with the main trainable model.

This is a prerequisite for implementing Direct Preference Optimization (DPO) #209 with normalization and for distillation-style training #214. The reference model must be completely frozen (no gradients, no optimizer), support parallelism (data, tensor, etc.), and be configurable independently from the primary model.

# 🚀 **Execution Plan**

### **Step 1: What is the smallest working version?**

- Extend the Fast-LLM training stack to accept an optional `reference_model` config block:
  - Separate architecture and checkpoint path
  - May use HuggingFace or pre-converted checkpoints
- Load the reference model at the beginning of training:
  - With the same distributed settings as the target model (the one being trained)
  - Distribute it across the same GPUs
- Ensure reference model:
  - Runs in `torch.no_grad()` mode
  - Never receives gradients or enters optimizer step
  - Uses same tokenizer as main model and has the same vocabulary size.
- At each training step:
  - Forward the input batch through the reference model
  - Expose its logits by providing a utility function to access ref model outputs for loss computation (e.g. `get_ref_logits(batch)`)

### **Step 2: What additional optimizations are possible (but optional)?**

- Allow reference model to use a subset of available GPUs instead of the same ones
- Allow reference model to use a different distributed/parallelism setup as the main model

# 📌 **Acceptance Criteria**

- Reference model can be configured and loaded independently
- Reference model runs in forward pass alongside main model
- Reference model is fully frozen (no optimizer, no gradients)
- Reference model outputs are usable in custom loss functions
- Training with a reference model does not break pretraining or SFT workflows
- Documented with examples for both DPO and distillation use cases

# 🛠️ **Project Management**

- [x] **Assign the project to the Fast-LLM project.**
- [ ] **Set the `Estimate` field (in days) in the GitHub project.**
- [x] **Use the `Size` field to categorize the PR size (Large).**
- [x] **Assign an owner when opening the issue.**

jlamypoirier:
Plan looks reasonable, implementation should be quite straightforward. Some comments:

> * Extend the Fast-LLM training stack to accept an optional `reference_model` config block:
>   
>   * Separate architecture and checkpoint path
>   * May use HuggingFace or pre-converted checkpoints
> * Load the reference model at the beginning of training:
>   
[...]

Issue #214, opened 2025-03-30
End-to-end knowledge distillation


tscholak:
# 🎯 **Goal (What & Why)**
Implement end-to-end knowledge distillation support, enabling training of a student model to mimic the outputs (logits) of a frozen teacher model.

# 🚀 **Execution Plan**

### **Step 1: What is the smallest working version?**
- Use existing support for loading a frozen reference (teacher) model from [Fast-LLM Issue #212](https://github.com/ServiceNow/Fast-LLM/issues/212).
- Add a knowledge distillation loss component (cross-entropy or KL-divergence) that compares logits from the teacher and student models.
- During training:
  - Compute logits from both the student and the frozen reference model.
  - Calculate and minimize the distillation loss.
  - Ensure the teacher model remains frozen and gradient-free.

### **Step 2: What additional optimizations are possible (but optional)?**
- tbd

# 📌 **Acceptance Criteria** (Must-Haves for Completion)
- The distillation loss is integrated and functional.
- Teacher model logits can be seamlessly accessed during student training.
- Tested end-to-end distillation workflow is documented clearly.
- Benchmark demonstrating accuracy retention after distillation.

# 🛠️ **Project Management**
- [x] **Assign the project to the Fast-LLM project.**
- [ ] **Set the `Estimate` field (in days) in the GitHub project.**
- [x] **Use the `Size` field to categorize the PR size (Small/Medium/Large).**
- [ ] **Assign an owner when opening the issue.**

nitsanluke:
Additional pointers from [NeMo distillation implementation](https://github.com/NVIDIA/NeMo/blob/main/examples/nlp/language_modeling/megatron_gpt_distillation.py).

[KL Loss](https://github.com/NVIDIA/NeMo/blob/edc87f0e6b36257960dd76a75e93610f55cb9dc6/examples/nlp/language_modeling/megatron_gpt_distillation.py#L342)
This makes use of ` modelopt.torch.distil` models for the trainer loop (not essential for us at this point)

Issue #209, opened 2025-03-26
Direct Preference Optimization (DPO) support


tscholak:
# 🎯 **Goal (What & Why)**

Add support for Direct Preference Optimization (DPO). This enables fine-tuning with preference-labeled data using a stable, RL-free objective. DPO guides the model to prefer `chosen` completions over `rejected` ones for the same prompt. This will broaden Fast-LLM's training capabilities while preserving a unified dataset structure across pretraining, SFT, and DPO.

# 🚀 **Execution Plan**

### Step 1: Implement simplified DPO without reference model

- Extend the `GPTSample` dataclass with two optional string fields:
  ```python
  chosen: str | None = None
  rejected: str | None = None
  ```
  i.e.
  | Field                 | Pretraining | SFT          | DPO (new)      |
  |----------------------|-------------|--------------|----------------|
  | `text`               | ✅ main input | ✅ main input | ✅ used as prompt |
  | `loss_masking_spans` | ❌           | ✅ used       | ❌ ignored      |
  | `chosen` / `rejected`| ❌           | ❌            | ✅ required     |

- Update `GPTMemmapDataset.__getitem__` to return `GPTSample` instances with `chosen` and `rejected` fields populated when present in the data.
- Modify `fast-llm prepare` to handle DPO-style preference data:
  - Add CLI flags to `fast-llm prepare` to force DPO mode or auto-detect based on fields.
  - Expect fields `text`, `chosen`, `rejected` (names configurable) in the input dataset.
  - Store all three fields in the bin files for training-time access.
- Add logic to ignore `loss_masking_spans` in DPO mode.
- Add config options for DPO, including β (the temperature).
- Implement DPO loss (simplified form):
  
  `L = -log σ(β * log(p_theta(chosen | x) / p_theta(rejected | x)))`

  where `p_theta` is the trained model and `x` is the prompt (text).

### Design options

**Packing**
1. Pack chosen, rejected pair from a single `GPTSample` into a sequence (similar to [OpenRLHF](https://github.com/OpenRLHF/OpenRLHF))
2. Pack multiple chosen/rejected pairs samples into a sequence
3. Distinct sequences for chosen and rejected samples.

Preferred option is to go with 2., but if it gets too complicated we can start with 1. Either methods would require splitting the packed sequence using `bos` token for getting the chosen and rejected logprobs.

We would need to modify `GPTSampledIndexedDataset`. It may also be easier to create a new class for DPO with similar logic as used for padding.
---

[...]

PR #216, opened 2025-04-01
Reference model support for distillation,. etc.

jlamypoirier:
# ✨ Description

Fix:#212

Add support for a reference model that can be used for distillation, etc.
* Config is  specified as `reference_models: dict[str, PretrainedFastLLMModelConfig]`. We probably won't need more than one such model, but this reduces config bloat in the default case.
* Model is run as a data preprocessor, and add its output to kwargs.
* Keeping the hack that sends the logits through kwargs, but we'll want something better eventually.
* Extracted an `InferenceRunner` class from the Hugging Face wrapper that can be used for Fast-LLM inference without the need to go through Hugging Face.
* Some tweaks to `Distributed` to deal with the distributed config being copied in the reference model.

It seems to be working and the tests pass, but we will need an actual use to properly test it.

```
/app$ fast-llm train gpt training.train_iters=10 data.format=random reference_models.teacher={}
2025-04-04 03:13:01,628 Using the legacy dataset definition format. Specify it through `data.datasets` instead.
2025-04-04 03:13:10,274 Setting random seeds...
2025-04-04 03:13:12,993 Creating model...
2025-04-04 03:13:13,031   Splitting the model into 14 stages...
2025-04-04 03:13:13,033   Total parameters: 181,543,936 
2025-04-04 03:13:13,034 Weight buffer placement:
{1: 0, 2: 1, 3: 2, 4: 3, 5: 4, 6: 5, 7: 6, 8: 7, 9: 8, 10: 9, 11: 10, 12: 11, 13: 12, 0: 13}
2025-04-04 03:13:13,034 Grad buffer placement:
{1: 0, 2: 1, 3: 2, 4: 3, 5: 4, 6: 5, 7: 6, 8: 7, 9: 8, 10: 9, 11: 10, 12: 11, 13: 12, 0: 13}
2025-04-04 03:13:13,034 Creating `teacher reference model...
2025-04-04 03:13:13,132   Splitting the model into 14 stages...
2025-04-04 03:13:13,134   Total parameters: 181,543,936 
2025-04-04 03:13:13,135 Weight buffer placement:
{1: 0, 2: 1, 3: 2, 4: 3, 5: 4, 6: 5, 7: 6, 8: 7, 9: 8, 10: 9, 11: 10, 12: 11, 13: 12, 0: 13}
2025-04-04 03:13:13,135 Grad buffer placement:
{1: 0, 2: 1, 3: 2, 4: 3, 5: 4, 6: 5, 7: 6, 8: 7, 9: 8, 10: 9, 11: 10, 12: 11, 13: 12, 0: 13}
2025-04-04 03:13:13,136 Command run:
/usr/local/bin/fast-llm train gpt training.train_iters=10 data.format=random 'reference_models.teacher={}'
2025-04-04 03:13:13,139 
------- fast_llm.models.gpt.config.GPTTrainerConfig --------
[...]
reference_models:
  teacher:
    model:
      [...]
[...]
--------------------------- end ----------------------------
2025-04-04 03:13:13,139 Setting up model...
2025-04-04 03:13:13,139 >>> Allocating 14 weight buffers (692.54 MiB)
2025-04-04 03:13:13,547 >>> Allocating 14 grad buffers (692.54 MiB)
2025-04-04 03:13:13,547 >>> Allocating 4 shards (2,770.14 MiB)
2025-04-04 03:13:13,548 Total allocated: 4,155.21 MiB
2025-04-04 03:13:13,553 Setting up `teacher` reference model...
2025-04-04 03:13:13,553 >>> Allocating 14 weight buffers (692.54 MiB)
2025-04-04 03:13:13,554 >>> Allocating 1 shards (692.54 MiB)
2025-04-04 03:13:13,554 Total allocated: 1,385.07 MiB
2025-04-04 03:13:13,568 Preparing datasets...
/app/fast_llm/data/data/gpt/data.py:99: UserWarning: The following datasets are defined but not used: validation, test. Ensure this is intentional, or update the configuration accordingly.
  warnings.warn(
2025-04-04 03:13:13,568 Preparing dataset. This may take several minutes.
/app/fast_llm/data/data/gpt/data.py:110: UserWarning: Using the dataset directory for the index cache.
  warnings.warn(f"Using the dataset directory for the index cache.")
2025-04-04 03:13:13,580 Initializing training state from scratch...
2025-04-04 03:13:13,917 No pretrained checkpoint specified for `teacher` reference model, using a freshly initialized model...
2025-04-04 03:13:13,931 done with setup ...
2025-04-04 03:13:13,932 After initial setup:  allocated 5,540.28 MiB | max allocated 5,540.28 MiB | reserved 5,550.00 MiB | max reserved 5,550.00 MiB | global max reserved 5,550.00 MiB
2025-04-04 03:13:13,932 Initializing training dataset iterator from sample 0...
2025-04-04 03:13:13,966 Training ...
2025-04-04 03:13:16,172 Data loading took 2,068.80 ms
```

Select all that apply:


fast_llm/engine/training/config.py
```diff
@@ -364,9 +375,24 @@ class TrainerConfig(PretrainedFastLLMModelConfig, ExperimentConfig):
...
 
     def _validate(self) -> None:
         self.training.export.setup(self.model)
+        self.model.validate()
+        if self.reference_models:
+            # TODO: Add support.
+            Assert.eq(self.model.distributed.pipeline_parallel, 1)
+            # TODO: Check if these work.
+            Assert.eq(self.model.distributed.tensor_parallel, 1)
+            Assert.eq(self.model.distributed.sequence_data_parallel, 1)
```
tscholak, line 392: It would be good to lift these restrictions because the reference models can and will be large, and the sequences will be long.

tscholak: is the distributed config shared between the student and the teacher models? if so, are we currently unable to use reference modes when training a model with pp, tp, sdp > 1?

fast_llm/engine/training/trainer.py
```diff
@@ -438,6 +453,19 @@ def _prepare_training_state(self) -> None:
...
+            if pretrained.path is not None and pretrained.model_weights:
+                log_main_rank(f"Loading weights for `{name}` reference model from {pretrained.path}")
+                reference_model.fast_llm_model.load_checkpoint(pretrained)
+            else:
+                log_main_rank(
+                    f"No pretrained checkpoint specified for `{name}` reference model,"
+                    f" using a freshly initialized model...",
+                    log_fn=logger.warning,
+                )
+                reference_model.fast_llm_model.initialize_weights()
```
tscholak, line 467: let's just error out here unless we are in debug mode. I don't see a single use case for this other than testing.

fast_llm/models/gpt/trainer.py
```diff
@@ -2,17 +2,35 @@
...
+    def preprocess_meta(self, kwargs: dict[str, typing.Any]) -> None:
+        pass
+
+    def preprocess(self, batch, kwargs: dict[str, typing.Any]) -> None:
+        # TODO: Fix random state/iteration.
+        preprocess_kwargs = kwargs.copy()
+        del preprocess_kwargs[LanguageModelKwargs.labels]
+        self._inference_runner.forward(batch, preprocess_kwargs, iteration=1)
+        # TODO: Improve.
+        kwargs[f"{self._name}_logits"] = preprocess_kwargs["logits"]
```
tscholak, line 29: This works well for now, since we'll need the logits downstream.

One small note: because the reference model has a user-defined name, the distillation loss in #214 and the DPO loss in #209 will need to use that name to access the logits. So the user-defined name should also be passed as an input to the loss functions, which complicates configuration.

tscholak (approved):
Thanks. This unblocks #214 and #209 with reference model.

Before we merge, can we address the limitations of this approach? I'm seeing `tp = 1`, `pp = 1`, and `sdp = 1` assertions, which gives me pause. In many cases, we'll want to distill from large models that won't fit on a single GPU, so FSDP alone won't cut it. Let's discuss how we can lift those constraints in a follow-up.

fast_llm/engine/training/config.py
jlamypoirier: PP would be really hard, but TP and SDP won't be that hard. The tricky part will be to make loss functions that use it compatible.

fast_llm/engine/training/trainer.py
jlamypoirier: We don't really have a "debug mode"...

fast_llm/models/gpt/trainer.py
jlamypoirier: Maybe not, we can probably get away with a hard-coded name or field default on the loss side, just like we enforce the training dataset to be called "training"

jlamypoirier:
> Before we merge, can we address the limitations of this approach? I'm seeing `tp = 1`, `pp = 1`, and `sdp = 1` assertions, which gives me pause. In many cases, we'll want to distill from large models that won't fit on a single GPU, so FSDP alone won't cut it. Let's discuss how we can lift those constraints in a follow-up.

This kind of inference doesn't take much memory, so FSDP alone should be enough for models well above 100B parameters.

tscholak:
> > Before we merge, can we address the limitations of this approach? I'm seeing `tp = 1`, `pp = 1`, and `sdp = 1` assertions, which gives me pause. In many cases, we'll want to distill from large models that won't fit on a single GPU, so FSDP alone won't cut it. Let's discuss how we can lift those constraints in a follow-up.
> 
> This kind of inference doesn't take much memory, so FSDP alone should be enough for models well above 100B parameters.

Ok, thanks, and I guess we are free to use any zero level here? Or are there limitations as well? I guess with zero 0 we wouldn’t be able to fit a 70B model, say.

Also, can you answer the question about whether the distributed configs of reference and student model are linked?

jlamypoirier:
> Ok, thanks, and I guess we are free to use any zero level here? Or are there limitations as well? I guess with zero 0 we wouldn’t be able to fit a 70B model, say.
>
> Also, can you answer the question about whether the distributed configs of reference and student model are linked?

The distributed configs are the same, at least for now. We can explore other options later, but having different configs would be really complicated.

We should be able to use any ZeRO level, as long as we're not doing generative inference. (Then ZeRO-3 is very inefficient.)
Note that the multi-stage config doesn't have to be the same, so we can use a different zero level for both.

PR #218, opened 2025-04-01
improvements to MTP implementation

RaymondLi0:
# ✨ Description

Follow-up to #179 with some fixes and improvements:
- [x] Add more labels in data-sampling to avoid input-truncation
- [x] handle sequence-parallelism and cross-entropy splits (as a consequence of the above)
- [x] Convert the additional lm-heads and add a corresponding hf-transformer implementation

Potential future work:
- [ ] add weights for each future-token prediction loss (e.g. the next token could be more important than the future ones)
- [ ] Improve throughput

## 🔍 Type of change

Select all that apply:


RaymondLi0:
Since there is no available implementation in the transformers library, I wrote a custom `modeling.py` for llama with multi-token prediction. 
The converter then copies these files in the export directory.
Wasn't sure what the best approach would be here, but that's the only thing I came up with. Let me know if there would be a better way.

fast_llm/layers/language_model/head.py
```diff
@@ -147,18 +144,23 @@ def _forward_backward(
     ) -> tuple[torch.Tensor, torch.Tensor | None]:
         labels = kwargs[LanguageModelKwargs.labels] if LanguageModelKwargs.labels in kwargs else None
         # MTP: Shift the labels
-        labels = labels[:, self._prediction_distance :].flatten() if labels is not None else None
+        labels = (
+            labels[
+                :,
+                self._prediction_distance : labels.shape[1]
```
jlamypoirier, line 150: `self._prediction_distance : self._prediction_distance + input_.size(1)`?

fast_llm/layers/language_model/head.py
```diff
@@ -60,6 +60,7 @@ def __init__(
         hidden_dim = self._tensor_space.get_tensor_dim(TransformerDimNames.hidden)
 
         self._loss_name = LanguageModelLossNames.multi_token_prediction_loss(prediction_distance)
+        self._logits_name = "logits" if prediction_distance == 0 else f"logits_{prediction_distance}"
```
jlamypoirier, line 63: Used only once, could be inline. Do we plan on combining knowledge distillation with MTP?

fast_llm/models/external/mtp_llama/modeling_mtp_llama.py
```diff
@@ -0,0 +1,963 @@
+from functools import partial
```
jlamypoirier, line 1: Is this mainly to allow benchmarks, etc.? We'll  have benchmarking directly in Fast-LLM soon, but I guess that's ok as a stopgap solution. 

Concerning the location, I don't think it goes there since it's not a model like the others. Maybe move inside `models/gpt`?

fast_llm/models/gpt/conversion.py
```diff
@@ -570,6 +584,110 @@ def _get_mlp_converters(self, fast_llm_prefix: str, hf_prefix: str) -> list[Weig
...
+
+    format: typing.ClassVar[type[CheckpointFormat]] = MTPLlamaGPTHuggingfaceCheckpointFormat
+    modeling_file = modeling_mtp_llama.__file__
+    configuration_file = configuration_mtp_llama.__file__
+    # modeling_file: typing.ClassVar[str] = "fast_llm/models/external/mtp_llama/modeling_mtp_llama.py"
+    # configuration_file: typing.ClassVar[str] = "fast_llm/models/external/mtp_llama/configuration_mtp_llama.py"
+    configuration_cls = MTPLlamaConfig
+
+    # Use custom config instead of relying on the transformers library
+    @classmethod
```
jlamypoirier, line 598: If we're planning on doing this again, might be worth extracting as a generic mixin. The structure looks already good for it.

fast_llm/data/data/gpt/data.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         vocab_size: int,
         max_sequence_length: int,
         cross_document_attention: bool = True,
+        prediction_heads: int = 1,
```
jlamypoirier, line 70: The dataset shouldn't have to worry about how it's used by the model, I suggest to rename to something more related to the present class. Maybe `extra_tokens`?

jlamypoirier (commented):
Looks good, some minor comments

fast_llm/layers/language_model/head.py
RaymondLi0: Much simpler indeed!

fast_llm/models/external/mtp_llama/modeling_mtp_llama.py
RaymondLi0: To allow benchmarks yes, but also other custom things like measuring the acceptance rate, etc.

Would you suggest to move it to `models/gpt/external/mtp_llama/modeling_mtp_llama.py` then?

fast_llm/layers/language_model/head.py
RaymondLi0: There is no short-term plan to combine knowledge distillation with MTP. Is there an issue supporting both though?

jlamypoirier: I don't think it will be hard, but we need to ensure the number of prediction heads is compatible (teacher>=student?)

fast_llm/models/external/mtp_llama/modeling_mtp_llama.py
jlamypoirier: Yes.

tscholak (approved):
LGTM!

jlamypoirier (approved):
Looks great, thanks!

PR #221, opened 2025-04-03
Fix sampling, add timeouts for test suprocess and data loaders

jlamypoirier:
# ✨ Description

#186 Broke sampling because of an incorrect walrus which always set `unshuffled_tokens` to 1 (`True`). I fixed it and simplified related code.

This bug only showed in slow tests (@sohamparikh please make sure these pass before merging) which didn't terminate because of a data loader crash somehow not being caught. I added some timeouts to ensure this doesn't happen again.

## 🔍 Type of change

Select all that apply:

### DIFF
--- a/fast_llm/data/dataset/gpt/sampled.py
+++ b/fast_llm/data/dataset/gpt/sampled.py
@@ -255,21 +255,20 @@ class GPTSampledIndexedDataset(SampledDataset):
         # Using `TOKEN_CUMSUM_RATE > 1` reduces pre-computation overhead at the cost of runtime computation.
         # Equivalent to `torch.hstack((0, document_sizes[all_document_index].cumsum()[::TOKEN_CUMSUM_RATE]))`
         if unshuffled_epochs > 0:
-            token_cumsum_unshuffled, num_tokens_unshuffled = self._get_token_cumsum(
+            token_cumsum_unshuffled, self._unshuffled_tokens = self._get_token_cumsum(
                 document_sizes,
                 offset=0,
                 # TODO: Allowing for max 100% extra tokens for padding, is that enough?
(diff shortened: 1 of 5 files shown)


fast_llm/data/dataset/gpt/sampled.py
```diff
@@ -432,10 +431,8 @@ def _lazy_load(self):
 
     def _load_yaml_data(self, data: dict[str, typing.Any]) -> None:
         self._documents_per_epoch = data["dataset"]["documents_per_epoch"]
-        if unshuffled_tokens := data.get("unshuffled_tokens") is not None:
-            self._unshuffled_tokens = unshuffled_tokens
-        else:
-            self._unshuffled_tokens = data["unshuffled_epochs"] * data["dataset"]["tokens_per_epoch"]
+        if "unshuffled_tokens" in data:
+            self._unshuffled_tokens = data["unshuffled_tokens"]
```
sohamparikh, line 435: should we care about backwards compatibility?

We could throw an error if `unshuffled_tokens` is not present in `loaded_yaml_data`. And speaking of backwards compatibility, I think we should also add `truncate_documents: True` in `loaded_yaml_data` if not present, else it will fail loading previous datasets.

sohamparikh (approved):
Thanks for catching it, LGTM! Just one comment (above) about backwards compatibility

fast_llm/data/dataset/gpt/sampled.py
jlamypoirier: The `if` was because of the method being called before `"unshuffled_tokens"` is added to data, I reorganized things so it's not needed.
We don't really care about backward compatibility for dataset cache, but I added a simple one.

Issue #217, opened 2025-04-01
Implement  generate with the mixin from HF 


tscholak:
# 🎯 **Goal (What & Why)**
**Implement `generate()` for `HuggingfaceGPTModelForCausalLM` using `GenerationMixin`, supporting greedy decoding only.**  
This makes the Fast-LLM model behave like a HuggingFace model for generation. The goal is to enable validation-time text generation directly from the sharded Fast-LLM model in memory, without converting to HF format (which would require extra memory and lead to model eviction).  
We use batched greedy decoding and support FlashAttention by padding and attention masking.  
No beam search, sampling, or KV caching is needed.

# 🚀 **Execution Plan**
Develop a minimal, batched, greedy generation loop using Fast-LLM's `.forward()` and `GenerationMixin` integration.

### **Step 1: What is the smallest working version?**
- Implement the `GenerationMixin` interface in `HuggingfaceGPTModelForCausalLM`, i.e. this interface:
    ```python
    class GenerationMixin:
        def prepare_inputs_for_generation(
            self,
            input_ids: torch.LongTensor,
            attention_mask: Optional[torch.LongTensor] = None,
            **kwargs,
        ):
            ...

        @torch.no_grad()
        def generate(
            self,
            inputs: None | torch.Tensor = None,
            generation_config: None | GenerationConfig = None,
            **kwargs,
        ) -> Union[GenerateOutput, torch.LongTensor]:
            ...
    ```
- Implement `generate()` to:
  - Accept `inputs` and `max_new_tokens`, `eos_token_id`, as well as `pad_token_id` via `GenerationConfig` or `kwargs`.
  - Only add support for batched greedy decoding (no sampling, no beam search).
  - Left-pad input sequences for FlashAttention.
  - Use an attention mask to exclude padding tokens.
  - Track completion using `eos_token_id` and `max_new_tokens`.
  - Reuse `.forward()` on the Fast-LLM model in each step.
- Implement `prepare_inputs_for_generation()` minimally to satisfy HF's expectations.
- Use the tokenizer to:
  - Tokenize prompts to input_ids.
  - Detokenize generated tokens to strings.
- Add a slow integration test that:
  - Loads `HuggingFaceTB/SmolLM2-135M-Instruct` in both HF and Fast-LLM formats.
  - Runs `.generate()` from both and compare outputs.
  - Use fixed prompt + seed + no sampling to make it deterministic.

### **Step 2: What additional optimizations are possible (but optional)?**
[...]

PR #222, opened 2025-04-03
Implementation of generate

bigximik:
# ✨ Description

part of #217 

Closes # 

## 🔍 Type of change

Select all that apply:

## 📝 Changes

1. Change A
2. Change B

## ✅ Checklist

### General

- [ ] 📜 I have read and followed the [contributing guidelines](https://servicenow.github.io/Fast-LLM/developers/contributing).

### Testing

### Performance Impact

---

## 🗒️ Additional Notes

### DIFF
--- a/fast_llm/engine/huggingface/model.py
+++ b/fast_llm/engine/huggingface/model.py
@@ -57,6 +57,7 @@ class HuggingfacePreTrainedModel(transformers.PreTrainedModel):
         )
         with transformers.modeling_utils.no_init_weights():
             self.post_init()
+        self._use_fm_changes = False
 
     @classmethod
     def from_pretrained(
@@ -78,13 +79,25 @@ class HuggingfacePreTrainedModel(transformers.PreTrainedModel):
         if torch_dtype is not None:
             config_updates[("distributed", "training_dtype")] = torch_dtype
 
+        attn_implementation = kwargs.pop("attn_implementation", None)
+        if attn_implementation is not None:
+            if attn_implementation == "flash_attention_2":
+                config_updates[("base_model", "transformer", "use_flash_attention")] = True
+            else:
+                config_updates[("base_model", "transformer", "use_flash_attention")] = False
+
         # Create the model
         fast_llm_model = cls.model_class.from_pretrained(
             pretrained_model_name_or_path, config_updates=config_updates, mode=mode
         )
         config = cls.config_class(fast_llm_model.config)
 
-        return cls(config, fast_llm_model, **kwargs)
+        obj = cls(config, fast_llm_model, **kwargs)
+
+        use_fm_changes = kwargs.pop("use_fm_changes", False)
+        obj._use_fm_changes = use_fm_changes
+
+        return obj
 
     def _init_weights(self, module) -> None:
         raise NotImplementedError(module)
--- a/fast_llm/models/gpt/huggingface.py
+++ b/fast_llm/models/gpt/huggingface.py
@@ -3,6 +3,7 @@ import random
 
 import torch
 import transformers.modeling_outputs
+import transformers.generation.utils as tgu
 
 from fast_llm.data.data.gpt.data import GPTBatch
 from fast_llm.engine.distributed.config import PhaseType
@@ -21,7 +22,7 @@ class HuggingfaceGPTModelConfig(HuggingfaceModelConfig):
     fast_llm_config: GPTModelConfig
 
 
-class HuggingfaceGPTModelForCausalLM(HuggingfacePreTrainedModel):
+class HuggingfaceGPTModelForCausalLM(HuggingfacePreTrainedModel, tgu.GenerationMixin):
     config_class = HuggingfaceGPTModelConfig
     config: HuggingfaceGPTModelConfig
     model_class = GPTModel
@@ -56,19 +57,38 @@ class HuggingfaceGPTModelForCausalLM(HuggingfacePreTrainedModel):
             raise NotImplementedError()
         if output_hidden_states:
             raise NotImplementedError()
-        if attention_mask is not None:
-            raise NotImplementedError()
-        if position_ids is not None:
-            raise NotImplementedError()
+        # if attention_mask is not None:
+        #     print(attention_mask)
+        #     raise NotImplementedError()
+        # if position_ids is not None:
+        #     print(position_ids)
+        #     raise NotImplementedError()
         if inputs_embeds is not None:
             raise NotImplementedError()
         if labels is not None:
             raise NotImplementedError()
 
+        if self._use_fm_changes:
+
+            # First non zero indexes or zero index if the row is all zeros (invalid row)
+            first_non_zero_indexes = attention_mask.argmax(dim=1)
+
+            # Check if the sequence is left-padded and if the remaining ones are continuous 1-ns
+            assert (attention_mask.sum(axis=1) == (attention_mask.shape[1] - first_non_zero_indexes)).all()
+
+            sequence_lenghts = [
+                torch.tensor(
+                    [attention_mask.shape[1]] if el == 0 else [el, attention_mask.shape[1] - el], dtype=torch.int64
+                )
+                for el in first_non_zero_indexes.tolist()
+            ]
+        else:
+            sequence_lenghts = None
+
         # Iteration serves as a random seed, using random module because it's not seeded by Fast LLM
         iteration = random.randint(0, 2**32)
         batch = self._fast_llm_model.base_model.preprocess(
-            GPTBatch(input_ids), phase=PhaseType.inference, iteration=iteration
+            GPTBatch(input_ids, sequence_lengths=sequence_lenghts), phase=PhaseType.inference, iteration=iteration
         )
         ((_, kwargs),) = batch
 
@@ -97,7 +117,7 @@ class HuggingfaceGPTModelForCausalLM(HuggingfacePreTrainedModel):
             past_key_values=kwargs[TransformerKwargs.presents],
         )
 
-    def prepare_inputs_for_generation(
-        self, input_ids, past_key_values=None, attention_mask=None, inputs_embeds=None, **kwargs
-    ):
-        raise NotImplementedError()
+    # def prepare_inputs_for_generation(
+    #     self, input_ids, past_key_values=None, attention_mask=None, inputs_embeds=None, **kwargs
+    # ):
+    #     raise NotImplementedError()
--- a/test.py
+++ b/test.py
@@ -1,0 +1,85 @@
+from transformers import AutoTokenizer, AutoModelForCausalLM
+from fast_llm.models.gpt.huggingface import HuggingfaceGPTModelForCausalLM
+from fast_llm.engine.checkpoint.config import CheckpointLoadConfig
+from fast_llm.models.gpt.config import LlamaGPTHuggingfaceCheckpointFormat
+
+import torch
+
+
+def diff_flm_hf(tokenizer, flm_tokens, hf_tokens):
+    print("+++++++++++++++fast_llm:+++++++++++++++++++++++++++++++++++++++++++++++++")
+    fllm_str = tokenizer.decode(flm_tokens)
+    print(fllm_str)
+    print("---------------hugging_face:---------------------------------------------")
+    hf_str = tokenizer.decode(hf_tokens)
+    print(hf_str)
+    print(
+        f"==============================({"Same" if fllm_str==hf_str else "Different"})====================================="
+    )
+
+
+def run_test(attn_implementation, torch_dtype, is_batch_size2, use_fm_changes):
+    checkpoint = "/mnt/checkpoints/pretrained_models/SmolLM2-135M-Instruct"
+
+    device = "cuda"  # for GPU usage or "cpu" for CPU usage
+    tokenizer = AutoTokenizer.from_pretrained(checkpoint)
+    # for multiple GPUs install accelerate and do `model = AutoModelForCausalLM.from_pretrained(checkpoint, device_map="auto")`
+    hf_kwards = {}
+    if attn_implementation is not None and attn_implementation == "flash_attention_2":
+        hf_kwards["attn_implementation"] = "flash_attention_2"
+    if torch_dtype is not None:
+        hf_kwards["torch_dtype"] = torch_dtype
+
+    print("hf_kwards", hf_kwards)
+    model = AutoModelForCausalLM.from_pretrained(checkpoint, **hf_kwards).to(device)
+
+    messages = [
+        {"role": "user", "content": "What is gravity?"},
+    ]
+    if is_batch_size2:
+        messages += [
+            {"role": "user", "content": "Who is the president of EU?"},
+        ]
+
+    input_text = [tokenizer.apply_chat_template([el], tokenize=False) for el in messages]
+
+    tokenizer.padding_side = "left"
+    inputs = tokenizer(input_text, padding="longest", return_tensors="pt").to(device)
+
+    outputs_hf = model.generate(**inputs, max_new_tokens=50, use_cache=False)
+    # print(tokenizer.decode(outputs_hf[0]))
+
+    fm_kwards = {}
+    if attn_implementation is not None and attn_implementation == "flash_attention_2":
+        fm_kwards["attn_implementation"] = "flash_attention_2"
+    else:
+        fm_kwards["attn_implementation"] = "fuse"
+    if torch_dtype is not None and torch_dtype == torch.bfloat16:
+        fm_kwards["torch_dtype"] = "bf16"
+
+    print("fm_kwards", fm_kwards)
+
+    fm_model = HuggingfaceGPTModelForCausalLM.from_pretrained(
+        CheckpointLoadConfig(
+            path=checkpoint,
+            format=LlamaGPTHuggingfaceCheckpointFormat,
+        ),
+        use_fm_changes=use_fm_changes,
+        **fm_kwards,
+    )
+
+    outputs = fm_model.generate(**inputs, max_new_tokens=50, use_cache=False)
+
+    diff_flm_hf(tokenizer, outputs[0][inputs["input_ids"].shape[1] :], outputs_hf[0][inputs["input_ids"].shape[1] :])
+    if len(outputs) > 1:
+        diff_flm_hf(
+            tokenizer, outputs[1][inputs["input_ids"].shape[1] :], outputs_hf[1][inputs["input_ids"].shape[1] :]
+        )
+
+
+def main():
+    run_test(attn_implementation=None, torch_dtype=None, is_batch_size2=False, use_fm_changes=False)
+
+
+if __name__ == "__main__":
+    main()


bigximik:
I have created a debugging sandbox with manual tests for now. The results are as follows:

## Ignoring `attention_mask` and `position_ids`:
| Batch Size | No Flash Attention (Float32) | No Flash Attention (BF16) | Flash Attention (BF16) |
|------------|------------------------------|---------------------------|------------------------|
| 1          | Same output                  | Same output               | Different output       |
| 2          | Different output             | Different output          | Different output       |

## Converting `attention_mask` (from HF `forward`) to `sequence_lengths`:
| Batch Size | No Flash Attention (Float32) | No Flash Attention (BF16) | Flash Attention (BF16) |
|------------|------------------------------|---------------------------|------------------------|
| 1          | FastLLM empty output         | FastLLM empty output      | Different output       |
| 2          | FastLLM empty output         | FastLLM empty output      | Different output       |

It seems `sequence_lengths` is not supported for fused attention and does not improve Flash Attention. Could this be correct?  

If `attention_mask` is a left-padded mask like this:  
`[[0, 0, 0, 1, 1, 1, 1], ....]`  
I convert it to `sequence_lengths = [[3, 4], ....]`.  
```python
# First non zero indexes or zero index if the row is all zeros (invalid row)
first_non_zero_indexes = attention_mask.argmax(dim=1)

# Check if the sequence is left-padded and if the remaining ones are continuous 1-ns
assert (attention_mask.sum(axis=1) == (attention_mask.shape[1] - first_non_zero_indexes)).all()

sequence_lenghts = [
    torch.tensor(
        [attention_mask.shape[1]] if el == 0 else [el, attention_mask.shape[1] - el], dtype=torch.int64
    )
    for el in first_non_zero_indexes.tolist()
]
```

bigximik:
@sohamparikh @jlamypoirier Hi, I am trying to use the cross-document attention prevention that @tscholak pointed me to (https://github.com/ServiceNow/Fast-LLM/pull/177/files) to mimic left padding for documents in a batch during generation. It appears to be doing the correct thing, such as building the internal mask and position IDs, but it is not working. Could you please comment on what might be wrong? Thanks!

fast_llm/engine/training/evaluator.py
```diff
@@ -0,0 +1,408 @@
...
+        self._data = data
+
+    @abc.abstractmethod
+    def run(
+        self,
+        done: bool,
+        completed_steps: int,
+        consumed_samples: int,
+        consumed_tokens: int,
+    ) -> tuple[dict[str, any], str | None]: ...
```
tscholak, line 81: use dataclass

fast_llm/engine/training/evaluator.py
```diff
@@ -0,0 +1,408 @@
...
+                total_losses[name] += value
+            self._run.save_logged_tensors(f"{full_phase_name}_{completed_steps}_{iter_}")
+
+        safe_barrier(
+            self._distributed.world_group,
+            f"{full_phase_name} end",
+        )
+        end_time = time.perf_counter()
+        time_per_iteration = (end_time - begin_time) / num_iters
+        model_tflops, hardware_tflops = self._get_tflops_func(phase, time_per_iteration)
```
tscholak, line 210: move downstream

fast_llm/engine/training/evaluator.py
```diff
@@ -0,0 +1,408 @@
...
+        eval_config: EvaluationLossConfig,
+        trainer_config: TrainerConfig,
+        get_tflops_func: callable,
+    ) -> "Evaluation":
+        return cls(
+            name=name,
+            eval_config=eval_config,
+            trainer_config=trainer_config,
+            get_tflops_func=get_tflops_func,
+        )
```
tscholak, line 57: make dataclass fields

fast_llm/engine/training/evaluator.py
```diff
@@ -0,0 +1,408 @@
...
+        eval_config: EvaluationLossConfig,
+        trainer_config: TrainerConfig,
+        get_tflops_func: callable,
+    ):
+        self._name = name
+        self._eval_config = eval_config
+        self._trainer_config = trainer_config
+        self._get_tflops_func = get_tflops_func
+
+        self._loss_defs = self._multi_stage.base_model.loss_defs
```
tscholak, line 107: use __post_init__?

fast_llm/engine/training/lm_eval/utils.py
```diff
@@ -0,0 +1,544 @@
...
+    args = parse_eval_args(parser, cli_args)
+
+    # NOTE: all this args are set by fast_llm on the model directly or not used here
+    assert not args.wandb_args  # default empty string
+    assert not args.wandb_config_args  # default empty string
+    assert args.model == "hf"  # default value of 'hf'
+    assert not args.model_args  # default empty string
+    assert args.batch_size == 1  # default value of 1
+    assert args.max_batch_size is None
+    assert args.device is None
```
tscholak, line 341: make sure these are raised during config class validation

fast_llm/engine/training/lm_eval/utils.py
```diff
@@ -0,0 +1,544 @@
...
+    """
+    for action in parser._actions:
+        if action.dest != "help" and not action.const:
+            if action.type is None:
+                raise ValueError(f"Argument '{action.dest}' doesn't have a type specified.")
+            else:
+                continue
+
+
+def setup_parser() -> argparse.ArgumentParser:
```
tscholak, line 75: use https://github.com/EleutherAI/lm-evaluation-harness/blob/e4a7b69fe0fc6cb430e12cf15c4109bf28185124/lm_eval/__main__.py#L83

fast_llm/engine/training/lm_eval/utils.py
```diff
@@ -0,0 +1,544 @@
...
+    #     wandb_logger = WandbLogger(wandb_args_dict, wandb_config_args_dict)
+
+    # TODO: change logging levels from fast_llm to lm_eval and then back?
+    # utils.setup_logging(args.verbosity)
+    # eval_logger = logging.getLogger(__name__)
+
+    os.environ["TOKENIZERS_PARALLELISM"] = "false"
+
+    # update the evaluation tracker args with the output path and the HF token
+    if args.output_path:
```
tscholak, line 354: please clean this up vvv
all this should be controlled by fast-llm

fast_llm/engine/training/lm_eval/utils.py
```diff
@@ -0,0 +1,544 @@
...
+    # if args.wandb_args:
+    #     wandb_args_dict = simple_parse_args_string(args.wandb_args)
+    #     wandb_config_args_dict = simple_parse_args_string(args.wandb_config_args)
+    #     wandb_logger = WandbLogger(wandb_args_dict, wandb_config_args_dict)
+
+    # TODO: change logging levels from fast_llm to lm_eval and then back?
+    # utils.setup_logging(args.verbosity)
+    # eval_logger = logging.getLogger(__name__)
+
+    os.environ["TOKENIZERS_PARALLELISM"] = "false"
```
tscholak, line 351: doesn't apply

jlamypoirier:
Can we please break down this PR? Otherwise it will make reviewing too difficult. Let's keep this one about the minimalistic `generate`, and move the rest to the next PR

tscholak:
> Can we please break down this PR? Otherwise it will make reviewing too difficult. Let's keep this one about the minimalistic `generate`, and move the rest to the next PR

Sure, eventually we can do that. @bigximik is currently iterating towards an end-to-end solution for running benchmarks, and he's solving issues as they arise. It makes sense for him to operate that way for the time being, but when the time comes to review the changes, we should separate the concerns.

tscholak:
@jlamypoirier, btw, we need your guidance in determining the best way to distribute generation across ranks.
Concretely, we are looking to implement this lm-eval-harness API:

```
    @abc.abstractmethod
    def generate_until(self, requests) -> List[str]:
        """Generate greedily until a stopping sequence

        :param requests: list[Instance]
            A list of Instance objects with property `args` which returns a tuple (context, gen_kwargs).
            context: str
                Context string
            gen_kwargs: dict
                A dictionary of keyword arguments to pass to the generation function e.g. top_k, until, etc.
        :return: list[str]
            A list of model generated continuations.
            continuation: str
                The generated continuation.
        """
        pass
```
where `generate_until(requests: list[Instance], ...)` is called from rank 0 and distribute the `Instance`s across ranks calling the Fast-LLM model's `generate(inputs: torch.Tensor, ...)`. An `Instance` is a prompt with fluff, https://github.com/EleutherAI/lm-evaluation-harness/blob/e4a7b69fe0fc6cb430e12cf15c4109bf28185124/lm_eval/api/instance.py#L11.

bigximik:
## Current State

- Implemented evaluation abstraction and `lm_eval` integration for single GPU.
- Made necessary changes to `generate()`.

## Next Steps

- Refactor `lm_eval` integration to rely less on moved code.
- Explore the possibility of using a base VLLM integration class instead of Hugging Face for `lm_eval`.
- Implement full distributed model support for `lm_eval` integration, including necessary changes to support distributed `generate()`.

fast_llm/engine/training/evaluator.py
```diff
@@ -0,0 +1,419 @@
...
+    def run(
+        self,
+        done: bool,
+        completed_steps: int,
+        consumed_samples: int,
+        consumed_tokens: int,
+    ) -> tuple[dict[str, any], str | None]: ...
+
+    @abc.abstractmethod
+    def get_dataset_samples(self) -> tuple[str, int] | None:
```
jlamypoirier, line 84: Suggest splitting in two, making into properties

fast_llm/engine/training/evaluator.py
jlamypoirier, line 1: I don't think the evaluator mechanism should belong to training, it should have its own submodule (or maybe go to inference?)

fast_llm/engine/training/config.py
```diff
@@ -154,6 +155,75 @@ class WandbConfig(Config):
 
 @config_class()
 class EvaluationConfig(IntervalConfig):
```
jlamypoirier, line 157: It probably doesn't make sense anymore to inherit from `IntervalConfig` since we're looking at evaluation more generally than just for training. How about we aim for something like:
```
training:
  evaluators:
    - name: validation
      interval: [IntervalConfig]
      evaluator:
        type: validation
        dataset: validation
```

fast_llm/engine/training/lm_eval/evaluator.py
```diff
@@ -0,0 +1,765 @@
...
+import logging
+import random
+import time
+from collections import defaultdict
+from typing import TYPE_CHECKING, List, Optional, Union
+
+import numpy as np
+import torch
+
+import lm_eval.api.metrics
```
jlamypoirier, line 12: Is that the present module or a separate package? Following our import conventions would help here.

fast_llm/engine/training/lm_eval/evaluator.py
```diff
@@ -0,0 +1,765 @@
+import itertools
+import json
+import logging
+import random
+import time
+from collections import defaultdict
+from typing import TYPE_CHECKING, List, Optional, Union
```
jlamypoirier, line 7: Please use the modern version of these for type hints,

fast_llm/engine/training/lm_eval/evaluator.py
```diff
@@ -0,0 +1,765 @@
...
+
+if TYPE_CHECKING:
+    from lm_eval.api.model import LM
+    from lm_eval.api.task import Task
+
+eval_logger = logging.getLogger(__name__)
+
+
+@positional_deprecated
+def simple_evaluate(
```
jlamypoirier, line 47: Is this just copy/paste from an external codebase?

fast_llm/engine/training/lm_eval/fast_llm_wrapper.py
```diff
@@ -0,0 +1,942 @@
...
+from fast_llm.engine.distributed.config import DistributedConfig
+
+from fast_llm.core.distributed import scatter_object_list, gather_object
+
+
+eval_logger = logging.getLogger(__name__)
+
+
+# move to fast_llm
+class FastLLMWrapper(TemplateLM):
```
jlamypoirier, line 42: Not clear what this does. Please use a more descriptive name

fast_llm/engine/training/lm_eval/utils.py
```diff
@@ -0,0 +1,544 @@
...
+    )
+    return parser
+
+
+def parse_eval_args(parser: argparse.ArgumentParser, args: list[str]) -> argparse.Namespace:
+    check_argument_types(parser)
+    return parser.parse_args(args)
+
+
+def prepare_lm_eval_simple_eval_params(
```
jlamypoirier, line 300: Why do we need this?

fast_llm/layers/transformer/preprocessing.py
```diff
@@ -239,7 +239,7 @@ def preprocess(self, batch, kwargs: dict[str, typing.Any]) -> None:
         ]
         if (sequence_lengths := kwargs.get(TransformerKwargs.sequence_lengths, None)) is not None:
             seq_ids = torch.stack(
-                [torch.cat([torch.arange(x) for x in sample_lens]) for sample_lens in sequence_lengths]
+                [torch.cat([torch.full((x,), i) for i, x in enumerate(sample_lens)]) for sample_lens in sequence_lengths]
```
jlamypoirier, line 242: Was this a bug?

jlamypoirier (commented):
Had a look, but the PR is too big to review properly. I suggest breaking down into smaller PRs, at the very least moving the `lm_eval` part to a separate PR. 
I also recommend adding some tests.

fast_llm/data/tokenizer.py
```diff
@@ -13,9 +13,18 @@ class Tokenizer:
 
     def __init__(self, config: TokenizerConfig):
         log_main_rank(f"> loading tokenizer from {config.path} ...")
-        self.tokenizer: PreTrainedTokenizerFast = PreTrainedTokenizerFast.from_pretrained(
-            pretrained_model_name_or_path=config.path, errors="replace", max_len=None
+        # self.tokenizer: PreTrainedTokenizerFast = PreTrainedTokenizerFast.from_pretrained(
+        #     pretrained_model_name_or_path=config.path, errors="replace", max_len=None
+        # )
+        self.tokenizer = AutoTokenizer.from_pretrained(
```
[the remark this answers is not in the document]
bigximik: Some models I'm using for testing—like Smol, if I'm correct—cannot instantiate their tokenizer from PreTrainedTokenizerFast, even though they do have a fast tokenizer.

fast_llm/engine/inference/config.py
```diff
@@ -91,7 +91,8 @@ def __eq__(self, other) -> bool:
 
     def to_dict(self) -> dict[str, typing.Any]:
         out = super().to_dict()
-        out["fast_llm_config"] = self.fast_llm_config.to_dict(verbose=FieldVerboseLevel.everything)
+        if self.fast_llm_config is not None:
```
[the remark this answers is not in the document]
bigximik: HF pre trained model constructor will try to serialize config, however fast_llm_config is not serializable on distributed run due to dims in distributed config having process groups assigned to them. so i reset and set it back during init

https://github.com/ServiceNow/Fast-LLM/blob/denis/generate/fast_llm/engine/inference/huggingface.py#L49

fast_llm/data/config.py
```diff
@@ -34,3 +34,8 @@ class TokenizerConfig(Config):
         desc="Path to the tokenizer file.",
         hint=FieldHint.core,
     )
+    bos_token: str | None = Field(
```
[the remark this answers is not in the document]
bigximik: Some models, like Qwen, do not have a BOS token, but we still need to set one because we use the same tokenizer for FIM and evaluations. We decide to use a separate tokenizer for evaluation, we'll still need a BOS token if we continue pretraining with FIM on models that don't have one.

fast_llm/engine/inference/huggingface.py
```diff
@@ -2,16 +2,22 @@
...
 from fast_llm.engine.multi_stage.config import StageMode
 from fast_llm.engine.multi_stage.fast_llm_model import FastLLMModel
+from fast_llm.engine.schedule.config import BatchConfig, ScheduleConfig
+from fast_llm.engine.schedule.runner import ScheduleRunner
+from fast_llm.engine.schedule.schedule import Schedule
+from fast_llm.engine.training.config import TrainerConfig
 
 
-class HuggingfacePreTrainedModel(transformers.PreTrainedModel):
+class HuggingfaceBaseModelForCausalLM(transformers.PreTrainedModel, transformers.generation.utils.GenerationMixin):
```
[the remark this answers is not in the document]
bigximik: We need a base wrapper class with a generate interface to pass to lm_eval and other potential integrations. So we decided that HuggingfaceBaseModelForCausalLM will serve as the base.

Please note that the GenerationMixin will be removed from the base PreTrainedModel class in the next version of Transformers. Also, we're not strictly following the Hugging Face design pattern here — if we were, a base model class like HuggingfacePreTrainedModel would be used inside the ForCausalLM model rather than inherited by it.

Maybe we should instead define a HuggingfacePreTrainedModelWithGenerate as the base?

fast_llm/engine/training/lm_eval/evaluator.py
bigximik: this will be removed as after decision on how lm_eval will be run just on rank 0 it is no longer needed to change this file

fast_llm/engine/training/lm_eval/evaluator.py
bigximik: this file will be removed

fast_llm/engine/training/lm_eval/fast_llm_wrapper.py
bigximik: FastLLMLmEvalWrapper, HuggingfaceBaseModelForCausalLMLMEvalWrapper?

fast_llm/engine/training/lm_eval/utils.py
bigximik: lm_eval integration accepts a list of lm_eval command line params like this:
```yaml
evaluations:
    gsm8k:
      type: lm_eval
      cli_args:
        - --tasks
        - gsm8k
        - --output_path
        - /mnt/checkpoints/test/denis/smol_eval_experiment/lm_eval
```

I implemented it this way because I’m not sure how stable their API is — for example, it has changed between version 0.4.8 and the current main branch.
So I decided to make our interface accept their command-line arguments and reuse their argument parser (with some modifications) to extract the parameters needed for internal functions like simple_evaluate.

This way, if the interface changes again, we’ll only need to adapt the parser. Ideally, backward compatibility will be handled by their parser itself.

fast_llm/layers/transformer/preprocessing.py
bigximik: yes

fast_llm/engine/inference/huggingface.py
```diff
@@ -54,18 +113,23 @@ def from_pretrained(
...
         )
-        config = cls.config_class(fast_llm_model.config)
 
+        config = cls.config_class(fast_llm_model.config)
         return cls(config, fast_llm_model, **kwargs)
 
     def _init_weights(self, module) -> None:
         raise NotImplementedError(module)
+
+    def can_generate(self):
```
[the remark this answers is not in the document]
bigximik: the mixin very weirdly check if it can actually generate so we decided with @tscholak  just to override the function to return true

fast_llm/engine/inference/runner.py
```diff
@@ -7,27 +7,47 @@
...
     batch_config_class: typing.ClassVar[type[BatchConfig]] = BatchConfig
 
-    def __init__(self, fast_llm_model: FastLLMModel):
+    def __init__(
+        self,
+        fast_llm_model: FastLLMModel,
+        trainer_config: TrainerConfig | None = None,
+        runner: ScheduleRunner | None = None,
+    ):
+        has_training_args = trainer_config is not None and runner is not None
```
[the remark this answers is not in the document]
bigximik: i have a class method for this:
https://github.com/ServiceNow/Fast-LLM/blob/denis/generate/fast_llm/engine/inference/huggingface.py#L91 which calls the constructor

how do you propose to do it other way?

check_logits_hidden_layers.ipynb
```diff
@@ -0,0 +1,400 @@
+{
```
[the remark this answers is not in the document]
bigximik: sorry

classes_fast_llm.jpg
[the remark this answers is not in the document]
bigximik: I just asked ChatGPT how to generate a class hierarchy — I think it was with Pyreverse.
We probably need to split the image by class groups.

Nice idea — I think we can do it during the doc build process.

fast_llm/engine/training/config.py
bigximik: Now it is like this:
```yaml
training:
  evaluations:
     name_of_validation_dataset: 
        interval: int
        offset: int
        iterations: int
```
It will be quite difficult to maintain backward compatibility if to change to evaluators and a list. Instead i left naming and dictionary and make things like this:
```yaml
training:
  evaluations:
    gsm8k:
      type: lm_eval
      cli_args:
        - --tasks
        - gsm8k
        - --output_path
        - /mnt/checkpoints/test/denis/smol_eval_experiment/lm_eval
    stack_3b:
       type: loss # optional for loss
       iterations: 10
       interval: 10
    fineweb:
       iterations: 10
       interval: 10
```

what do you think?

fast_llm/engine/training/lm_eval/evaluator.py
bigximik: the file will be removed

fast_llm/engine/training/evaluator.py
bigximik: yeh maybe inference is a good place is it is inference  one way or another, will move

fast_llm/engine/training/lm_eval/utils.py
bigximik: Moving this to _validate will require lm_eval be installed for a dry_run also to separate parameter preparation function, as it nows not just passes arguments, but also creates lm_eval tasks and evaluation_tracker objects

fast_llm/engine/training/lm_eval/utils.py
bigximik: done

fast_llm/engine/training/lm_eval/utils.py
bigximik: it is done for the variant of passing cli args for lm_eval integration, if we decide to implement our own params set for the lm_eval integration, it will be done separately

fast_llm/engine/training/lm_eval/utils.py
bigximik: done

fast_llm/engine/training/lm_eval/utils.py
bigximik: tasks are object which are actually passed to evaluate, and evaluation_tracker tracks and saves details to disk

fast_llm/engine/training/evaluator.py
```diff
@@ -0,0 +1,419 @@
...
+from fast_llm.engine.training.lm_eval.fast_llm_wrapper import FastLLMWrapper
+from fast_llm.engine.training.lm_eval.utils import prepare_lm_eval_simple_eval_params, process_lm_eval_results
+
+# from fast_llm.engine.training.lm_eval.evaluator import simple_evaluate as lm_eval_simple_evaluate
+from lm_eval.evaluator import simple_evaluate as lm_eval_simple_evaluate
+
+logger = logging.getLogger(__name__)
+
+
+class Evaluation[ConfigType: EvaluationConfig](Configurable[ConfigType], abc.ABC):
```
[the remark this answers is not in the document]
bigximik: done

fast_llm/engine/training/lm_eval/evaluator.py
bigximik: done

fast_llm/engine/training/lm_eval/evaluator.py
bigximik: done

fast_llm/engine/training/lm_eval/fast_llm_wrapper.py
bigximik: selected FastLLMLmEvalWrapper

fast_llm/core/distributed.py
```diff
@@ -133,3 +249,406 @@ def set_generator(generator: torch.Generator) -> typing.Generator[None, None, No
     finally:
         generator.set_state(default_generator.get_state())
         default_generator.set_state(old_state)
+
+
+def gather(
```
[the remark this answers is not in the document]
jlamypoirier: I see. These can be simplified to 5-6 lines like the ones above though, we don't need all the features and checks from the pytorch implementations because we already have standardized imputs.

fast_llm/tools/evaluate.py
```diff
@@ -0,0 +1,25 @@
+import argparse
```
jlamypoirier, line 1: I don't really the point of this, it's just a synonym of training. Let's remove and maybe later add proper standalone evaluation.

fast_llm/engine/training/config.py
jlamypoirier: One problem is that the interval and offset shouldn't be part of the evaluator config, It's needed for all evaluators during training but not for standalone evaluation. So they should be separate from the evaluation config in some way.

I agree a dict of evaluation is better than a list, but I'm not sure about using the key as the dataset name: do we really want to enforce a 1-1 match between datasets and evaluators?

fast_llm/tools/evaluate.py
bigximik: Having a separate cli command for evaluation — even if it's just a wrapper around training — was requested by @tscholak, so that people can start using it with our models with fml-ops.

fast_llm/engine/training/evaluator.py
bigximik: get_tflops moved to model and called from it, done

fast_llm/engine/training/evaluator.py
bigximik: configurables are not dataclasses, but @jlamypoirier  proposed for EvaluatorConfig to return Evaluator instance as means of interface abstraction.

fast_llm/engine/training/evaluator.py
bigximik: Evaluator is Configurable, so it has no __post_init__, so this is moved to setup

fast_llm/engine/training/evaluator.py
```diff
@@ -0,0 +1,419 @@
...
+    config_class: typing.ClassVar[type[EvaluationConfig]] = EvaluationConfig
+
+    _is_setup: bool = False
+
+    @classmethod
+    def build(
```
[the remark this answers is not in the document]
bigximik: done

fast_llm/engine/training/config.py
jlamypoirier: Simplest is some kind of `TrainingEvaluatorConfig` config that is a composition of an `EvaluatorConfig` and `IntervalConfig`, as my example above suggests.
Other option is to use multiple inheritance to associate a `TrainingEvaluatorConfig` to each `EvaluatorConfig` (ex. `class TrainingValidationConfig(ValidationConfig, IntervalConfig)` that would produce neater configs at the cost of extra code.

fast_llm/engine/inference/huggingface.py
```diff
@@ -20,31 +26,84 @@ class HuggingfacePreTrainedModel(transformers.PreTrainedModel):
...
+        use_cache: bool | None = None,
+        output_attentions: bool | None = None,
+        output_hidden_states: bool | None = None,
+        return_dict: bool | None = None,
+    ) -> tuple | transformers.modeling_outputs.CausalLMOutputWithPast:
+        # Meant to be overridden in derived classes
+        raise NotImplementedError()
+
+    @classmethod
+    def from_fast_llm_model_in_training(
```
jlamypoirier, line 91: `def from_trainer(cls, trainer: Trainer, **kwargs):`

fast_llm/engine/inference/runner.py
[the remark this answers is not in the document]
jlamypoirier: I guess this is ok, but let's pass a trainer here instead.

fast_llm/engine/inference/huggingface.py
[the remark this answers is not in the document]
jlamypoirier: Ok, let's add a short comment?

fast_llm/tools/evaluate.py
tscholak: Btw, I get that under the hood this could be routed through train with `train_iters=0`. Thanks for pointing this out, @jlamypoirier. I believe that's what @bigximik is doing here.
The point is that we expose a clear `evaluate` command to users. Even if it's starts out as a thin wrapper, that's a win. It gives us the simplicity and clarity we want. Perfect is the enemy of good enough. Let's land this and improve it incrementally if needed.

fast_llm/engine/inference/huggingface.py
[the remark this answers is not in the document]
jlamypoirier: We're not that far off actually, GPT task-specific classes all inherit from a model-specific base class (ex. `class GPT2LMHeadModel(GPT2PreTrainedModel, GenerationMixin):`). The point of `HuggingfacePreTrainedModel` is to make an equivalent to  the model-specific base class and/or `PreTrainedModel`.

Where we really differ from Hugging Face is that we wrap a whole model (runner) and don't separate the model head.

fast_llm/data/config.py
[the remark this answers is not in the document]
jlamypoirier: By separate tokenizer you mean a completely different one or the same with added bos?

fast_llm/engine/inference/config.py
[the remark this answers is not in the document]
jlamypoirier: It's not really in `DistributedConfig`, but `DistributedConfig` may have an indirect pointer to `Distributed` through  `DistributedDim`'s pointer to a `TensorSpace`. Anyway we can't let HF do this, but where is it done? Only deepcopy I see is in `from_pretrained`, and `__init__` doesn't go through that path.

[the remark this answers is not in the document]
bigximik: Here is a partial stack trace, it is inside generation_config creation:
```shell
File "/home/toolkit/code/Fast-LLM/fast_llm/engine/inference/huggingface.py", line 54, in __init__
    super().__init__(config, **kwargs)
  File "/home/toolkit/code/Fast-LLM/.venv/lib/python3.12/site-packages/transformers/modeling_utils.py", line 1884, in __init__
    self.generation_config = GenerationConfig.from_model_config(config) if self.can_generate() else None
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/toolkit/code/Fast-LLM/.venv/lib/python3.12/site-packages/transformers/generation/configuration_utils.py", line 1277, in from_model_config
    config_dict = model_config.to_dict()
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/toolkit/code/Fast-LLM/fast_llm/engine/inference/config.py", line 93, in to_dict
    out = super().to_dict()
          ^^^^^^^^^^^^^^^^^
  File "/home/toolkit/code/Fast-LLM/.venv/lib/python3.12/site-packages/transformers/configuration_utils.py", line 877, in to_dict
    output = copy.deepcopy(self.__dict__)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

fast_llm/data/config.py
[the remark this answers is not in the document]
bigximik: Yeah, right now the only tokenizer used is the one created inside the data object using our wrapper, which checks for the mandatory presence of a BOS token.

The BOS token is not needed for `lm_eval`, so if the `lm_eval` wrapper had its own tokenizer, we could avoid requiring the BOS token for models that don't have it. But this only applies to the `lm_eval` integration.

If we later perform FIM on a model that doesn't have a BOS token, we’ll still need to specify it manually.

fast_llm/engine/training/config.py
bigximik: I have made changes, now the config looks like this:
```yaml
evaluations:
    gsm8k:
      interval:
        interval: 10
      evaluator:
        type: lm_eval
        cli_args:
          - --tasks
          - gsm8k
          - --output_path
          - /mnt/checkpoints/test/denis/smol_eval_experiment/lm_eval
    stack_3b:
      interval:
        interval: 10
      evaluator:
        type: loss # optional for loss
        iterations: 10
        dataset_name: stack_3b # optional if omitted evaluation name will be used for dataset name
    fineweb:
      interval:
        interval: 10
      evaluator:
        iterations: 10
        dataset_name: fineweb
```

Also, `TrainingEvaluatorConfig` and `EvaluatorConfig`  and `TrainingEvaluator` and `Evaluator` have exactly the same interface. `TrainingEvaluator` just wraps around `Evaluator` to track being called multiple times during training. So, EvaluatorRunner can accept both dictionaries of configs transparently and is ready to be run with separate Evaluation Config

fast_llm/engine/inference/huggingface.py
bigximik: i have changed to `from_model` and it accepts optional `runner` and `micro_batch_size`

fast_llm/tools/evaluate.py
bigximik: Okay, I will leave `evaluate` as-is for now. Once [this PR](https://github.com/ServiceNow/Fast-LLM/pull/245) is merged, I’ll create a proper `evaluate` command that can accept either a `TrainerConfig` or an `EvaluationConfig`.

We'll likely need to call the config for the `evaluate` command `EvaluationConfig`, since `EvaluatorConfig` is already used for individual evaluations. Alternatively, we could name it `EvaluatorRunnerConfig`.

@jlamypoirier @tscholak — thoughts on the config naming?

fast_llm/engine/inference/runner.py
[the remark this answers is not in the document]
bigximik: i made it from_model with optional runner and micro_batch_size

fast_llm/engine/training/config.py
```diff
@@ -154,6 +155,75 @@ class WandbConfig(Config):
...
+                raise ValueError(
+                    f"Registry {cls._registry.name} already contains type {cls.type_}."
+                    f" Make sure all classes either have a unique or `None` type."
+                )
+            EvaluationConfig._registry[cls.type_] = cls
+        super().__init_subclass__()
+
+
+@config_class()
+class EvaluationLossConfig(EvaluationConfig):
```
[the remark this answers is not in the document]
bigximik: Have done all the renaming

fast_llm/engine/training/evaluator.py
bigximik: They are both need to be present or both to be absent, i will wrap return value in a small data class for clarity instead

fast_llm/engine/inference/huggingface.py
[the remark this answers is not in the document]
bigximik: Checked again default implementation works for our case so removed override.

check_logits_hidden_layers.ipynb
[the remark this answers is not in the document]
bigximik: done

classes_fast_llm.jpg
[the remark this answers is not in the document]
bigximik: removed

fast_llm/core/distributed.py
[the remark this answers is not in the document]
bigximik: Yeah, makes sense. But most of the lines are documentation, and type checking doesn’t take up much space. Also, while the probability of passing something incorrect to functions already used internally is low (since they’re inside layers), `scatter` and `gather` are passed data external to our library. So maybe we should leave the checks in as additional safeguards?

fast_llm/engine/inference/huggingface.py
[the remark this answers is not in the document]
bigximik: I will restore `HuggingfaceBaseModel`, but I’ll add a `HuggingfaceBaseModelForCausalLM`, from which `HuggingfaceGPTModelForCausalLM` and other causal LM variants will inherit.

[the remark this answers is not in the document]
bigximik: done

fast_llm/engine/multi_stage/stage.py
```diff
@@ -185,6 +197,16 @@ def invalidate_buffer(self) -> None:
         for fsdp in self._fsdps:
             fsdp.invalidate_buffer()
 
+    @torch._dynamo.disable  # noqa
+    def _get_global_output_tensor(
```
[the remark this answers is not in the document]
bigximik: done

fast_llm/engine/multi_stage/stage.py
```diff
@@ -111,6 +114,15 @@ def forward(
...
             )
             self._log_layer_forward(output, kwargs, i)
+
+            # TODO: very slow and memory consuming, only use for debugging for now
+            # TODO: decide if and how we want to return 
+            #       HF transformer style details from forward properly
+            if "output_hidden_states" in kwargs and kwargs["output_hidden_states"]:
+                kwargs["hidden_states"][self._layer_range[i]] = {
+                    "layer_type": type(layer).__name__,
+                    "tensor": self._get_global_output_tensor(i, output),
```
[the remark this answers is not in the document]
bigximik: done

fast_llm/engine/schedule/runner.py
```diff
@@ -396,8 +396,11 @@ def _recv(self, context: BatchContext, step: Step) -> None:
             self._record_event(context, EventType.compute_wait_pipe, step)
 
     def _forward(self, context: BatchContext, step: Step) -> None:
+        input = self._get_forward_input(context, step)
+        if not "hidden_states" in context.batch[step.data_index]:
+            context.batch[step.data_index]["hidden_states"] = {}
```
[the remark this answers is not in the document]
bigximik: done

fast_llm/engine/training/evaluator.py
bigximik: moved to engine/evaluation

bigximik:
I’ve finished working in this draft and will create 3 new PRs from it:

- Generate support  
- Refactoring of evaluations  
- `lm_eval` integration  

In addition to the changes here, I’ll be adding tests and documentation updates as needed.

I’ll also be tracking this draft in case further discussion continues here.

fast_llm/engine/inference/huggingface.py
jlamypoirier: `from_trainer` would be more appropriate since it's the construct we are actually using.

fast_llm/engine/training/config.py
```diff
@@ -152,22 +153,37 @@ class WandbConfig(Config):
...
-        default=None,
-        desc="Number of iterations for each evaluation phase. Setting to None will disable.",
-        hint=FieldHint.feature,
-        valid=skip_valid_if_none(check_field(Assert.gt, 0)),
-    )
-
-    def get_iteration_count(self, training_iterations: int, extra_evaluations: int = 0):
-        # Number of completed validation iterations
-        return (self.get_count(training_iterations) + extra_evaluations) * self.iterations if self.enabled() else 0
+class TrainingEvaluatorConfig(EvaluatorConfigBase):
```
jlamypoirier, line 156: I don't see the point of the inheritance. We can just call `.evaluator.get_evaluator()`

fast_llm/engine/training/config.py
```diff
@@ -152,22 +153,37 @@ class WandbConfig(Config):
...
+    run_interval: IntervalConfig = Field(default_factory=IntervalConfig)
+    evaluator: EvaluatorConfig = Field(default_factory=EvaluatorConfig)
+
+    def get_run_count(self, training_iterations: int, extra_evaluations: int = 0):
+        # Number of completed evaluation runs
+        return (self.run_interval.get_count(training_iterations) + extra_evaluations) if self.run_interval.enabled() else 0
+
+    def get_evaluator(
```
jlamypoirier, line 164: Why this?

fast_llm/engine/training/config.py
bigximik: so the evaluator runner does not need to know which one it is instantiating, it works just with the interface:

https://github.com/ServiceNow/Fast-LLM/blob/denis/generate/fast_llm/engine/evaluation/evaluator.py#L352

this is in preparation for the standalone evaluation command, actually all is ready for it we just need to decide how to name EvaluationConfig for it and how exactly to put needed fields, do we want to mimic Trainer config or just put all on the same base level?

fast_llm/engine/training/config.py
bigximik: In `TrainerConfig`, we have a `dict[str, TrainingEvaluatorConfig]`, which supports multiple evaluation runs (controlled via `run_interval`).

In `EvaluationConfig`, we'll have a `dict[str, EvaluatorConfig]`, which supports a single evaluation run.

Both types can be passed transparently to `EvaluatorRunner`, allowing it to handle either case uniformly.

bigximik:
Work on this prototype branch has been completed and moved to other feature branches. This PR can be safely closed.

PR #223, opened 2025-04-03
DPO

### LINKED ISSUE #209: Direct Preference Optimization (DPO) support

tobyzl2:
# Description

This PR introduces the implementation of Direct Preference Optimization (DPO) training on Fast-LLM. DPO enhances model fine-tuning by directly incorporating user preferences into the optimization process, ensuring that the model better aligns with the desired output behavior.

Closes #209 

Throughput Numbers (reference free - 04/14/25)
| Model         | Tok/Sec/GPU | Seq Len | Num Steps | # GPUs |
|---------------|-------------|---------|-----------|--------|
| Mistral 7b    | 10168.015   | 8192    | 400       | 32     |
| Apriel 5b SFT | 14722.243   | 8192    | 6000      | 32     |

Throughput Numbers (with reference model - 04/29/25)
| Model         | Tok/Sec/GPU | Seq Len | Num Steps | # GPUs |
|---------------|-------------|---------|-----------|--------|
| Mistral 7b    | 7837.249    | 8192    | 400       | 32     |
| Apriel 5b SFT | 11331.242   | 8192    | 6000      | 32     |

## 🔍 Type of change

Select all that apply:

## 📝 Changes

1. introduced DPO training with simplified DPO loss
2. allows users to configure DPO parameters (beta value)
3. allows option for packing to be turned off (currently dpo is implemented without packing)

## ✅ Checklist

### General

- [ ] 📜 I have read and followed the [contributing guidelines](https://servicenow.github.io/Fast-LLM/developers/contributing).

### Testing

### Performance Impact

---

## 🗒️ Additional Notes


fast_llm/layers/language_model/head.py
```diff
@@ -174,6 +181,38 @@ def _forward_backward(
             return loss, input_.grad
         else:
             return loss, None
+        
+    def _logits_dpo(
```
jlamypoirier, line 185: The logits computation looks unchanged, why re-implementing it? Seems like you just need to swap `cross_entropy_forward_backward` with `compute_simplified_dpo_loss` in `_logits_cross_entropy_forward_backward`

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -34,13 +34,17 @@ def _init(self, name: str, prefix: pathlib.Path | str, num_documents: int | None
...
+        self._has_preference_spans = False
 
         with self._prefix.with_suffix(".idx").open("rb") as stream:
             Assert.eq(stream.read(9), MEMMAP_INDEX_HEADER, msg=f"File: {stream.name}")
             self._version = struct.unpack("<Q", stream.read(8))[0]
-            assert self._version in [1, 2], f"Unsupported version for gpt_memmap dataset: {self._version}."
+            assert self._version in [1, 2, 3], f"Unsupported version for gpt_memmap dataset: {self._version}."
             if self._version == 2:
                 self._has_spans = struct.unpack("<B", stream.read(1))[0]
+            if self._version == 3:
```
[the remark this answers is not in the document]
tobyzl2: Yup!  Just added a test case here for v3 with preference pairs

https://github.com/ServiceNow/Fast-LLM/blob/0e0909850c0e152394eee077c1294ec94ad7d46a/tests/data/test_prepare_gpt_memmap.py#L41

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -83,6 +92,34 @@ def _init(self, name: str, prefix: pathlib.Path | str, num_documents: int | None
...
+            rejected_span_offset = offset + self._document_sizes.nbytes + self._pointers.nbytes + np.array(self._chosen_spans).nbytes
+            for idx in range(self._num_documents):
+                self._rejected_spans.append(
+                    np.frombuffer(
+                        self._index_bin_buffer,
+                        dtype=np.int32,
+                        count=2,
+                        offset=rejected_span_offset + idx * 2 * np.dtype(np.int32).itemsize,
+                    )
+                )
```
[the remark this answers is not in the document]
tobyzl2: Chosen/rejected spans should always have 2 values specifying the begin/end tokens for loss masking but the range of these spans can differ as chosen or rejected generations can have varying lengths.  We can maybe add a check just to make sure that these spans are in valid format though.

fast_llm/data/dataset/gpt/sampled.py
```diff
@@ -254,45 +274,70 @@ def _sample(self) -> None:
...
+                self._document_shuffling.save(
+                    document_shuffling[:self._num_samples].numpy(
+                        force=self._config.gpu
+                    )
+                )
+            self._document_sizes.save(document_sizes.numpy(force=self._config.gpu))
+            if self._yaml_path is not None:
+                # yaml_data["unshuffled_tokens"] = num_tokens_unshuffled
+                self._yaml_path.parent.mkdir(parents=True, exist_ok=True)
+                yaml.safe_dump(yaml_data, self._yaml_path.open("w"))
```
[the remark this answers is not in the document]
tobyzl2: I believe it's used for detecting if a sampled dataset is already cached and also used in lazy loading some values in the __get_item__ function.  The same code is done here currently, I've just added it again for the DPO case. 

https://github.com/ServiceNow/Fast-LLM/blob/edca385c0945bd276d1c8540c132235b3498835c/fast_llm/data/dataset/gpt/sampled.py#L293

fast_llm/functional/dpo.py
```diff
@@ -0,0 +1,51 @@
...
+    rejected_mask[rejected_span[:, 0]: rejected_span[:, 1] + 1] = 1
+    rejected_logp = (selected_log_probs * rejected_mask).sum()
+
+    return chosen_logp, rejected_logp
+
+def compute_simplified_dpo_loss(
+    logits: torch.Tensor, 
+    targets: torch.Tensor, 
+    chosen_span: torch.Tensor, 
+    rejected_span: torch.Tensor,
```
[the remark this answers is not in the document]
tobyzl2: Currently one span per document since we have to do DPO a bit differently from SFT.

The reason for this change is due to the multiturn scenario since the conversation history changes at each turn when we sample only the chosen turn for the history.  Because of this, this requires us to split the data ahead of time and compute the loss on only the last assistant turn.

For example for a preference sample with 2 turns, we have to unroll it into the following sequences since we are always using the chosen turn for the conv history.
Prompt1, **Chosen1**
Prompt1, **Rejected 1**
Prompt1, Chosen1, Prompt 2, **Chosen 2**
Prompt1, Chosen1, Prompt 2, **Rejected 2**

fast_llm/data/tokenizer.py
```diff
@@ -60,7 +65,11 @@ def tokenize_with_spans(
                 input_ids.extend(tokenized_text)
             curr_text = text[start : end + 1]
             if end >= len(text) - 1:
-                tokenized_text = self.tokenize(curr_text, begin=beginning_of_text, end=True)
+                tokenized_text = self.tokenize(
+                    curr_text, 
+                    begin=beginning_of_text, 
+                    end=True if not curr_text.endswith(self.eod_token) else False
```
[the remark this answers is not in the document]
sohamparikh: I think we should not expect text to already include bos/eos tokens, it would make data preparation less error prone. Once we add support for chat templates (need to create an issue), it should handle other kinds of special tags too, so I believe we should move towards that

fast_llm/functional/dpo.py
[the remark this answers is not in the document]
tscholak: I don’t have a strong opinion here, but I was confused that the loss is only taking one reject span and one accept span per document, whereas the rest of the codebase appear to support more than one.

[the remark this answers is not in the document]
tscholak: > For example for a preference sample with 2 turns, we have to unroll it into the following sequences since we are always using the chosen turn for the conv history. (Bold are the parts that we are applying loss mask)
> 
> Prompt1, **Chosen1** Prompt1, **Rejected 1** Prompt1, Chosen1, Prompt 2, **Chosen 2** Prompt1, Chosen1, Prompt 2, **Rejected 2**

I don't recall seeing any code in this PR doing this unrolling. Could you point to where this is happening?

fast_llm/data/dataset/gpt/memmap.py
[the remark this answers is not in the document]
tscholak: what I meant was `assert len(self._chosen_spans) == len(self._rejected_spans)`. does that make sense?

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -116,13 +153,47 @@ def get(
...
+
+            # subtract by offset to normalize span boundaries
+            sample_spans[:, 0] = np.maximum(sample_spans[:, 0], offset) - offset # offset 
             sample_spans[:, 1] = np.minimum(sample_spans[:, 1], offset + len(token_ids) - 1) - offset
-        return GPTSample(token_ids=token_ids, loss_masking_spans=sample_spans)
+        
+        chosen_spans = None
+        rejected_spans = None
+        if use_preference_loss_masking_spans and self._chosen_spans is not None and self._rejected_spans is not None:
+            chosen_spans = self._chosen_spans[idx]
```
tscholak, line 169:
```suggestion
            chosen_span = self._chosen_spans[idx]
```

fast_llm/layers/language_model/head.py
tobyzl2: Yes, makes sense since most of those computations can be shared.  I've moved the dpo loss call over into the `_logits_cross_entropy_forward_backward` function which will route between the two loss types.

fast_llm/data/tokenizer.py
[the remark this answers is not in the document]
tobyzl2: Sounds good, just made the change to revert this.

fast_llm/data/dataset/gpt/memmap.py
tobyzl2: Just did some refactoring, chosen_spans -> chosen_span, rejected_spans -> rejected_span throughout unless it is part of a batch.

fast_llm/functional/dpo.py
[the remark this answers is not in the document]
tobyzl2: Thanks for the suggestions, I've made the change so now instead of providing spans, users can provide the conversation history, chosen text, and rejected text and span computations will now be done within the `prepare` command.

For unrolling, we're currently assuming that this is done offline as part of data preparation though I believe this is would be a useful feature to have.  Currently, it's trickier to implement as the input is conversation history/chosen/rejected but I think it would be good to discuss this further as it would be a common need for multiturn training.

fast_llm/layers/language_model/head.py
```diff
@@ -287,14 +294,25 @@ def _logits_cross_entropy_forward_backward(
...
+                group=self._tensor_space.distributed.tensor_group if self._parallel_embeddings else None,
+                grad_output=grad_output,
+                implementation=self._cross_entropy_impl,
+                logits_scale_factor=self._logits_scale_factor,
+            )
+        elif self._loss_function_type == LossFunctionType.dpo:
+            loss, grad = compute_simplified_dpo_loss(
+                logits.flatten(0, -2),
+                labels,
+                kwargs[LanguageModelKwargs.chosen_spans],
```
jlamypoirier, line 310: How do we ensure it's there? Seems like this will crash unless:
* All datasets have both chosen and rejected spans.
* The sampling config for all datasets is set to use these spans.

It's a bad idea to wait so late for crash, we should aim to do the check sooner

fast_llm/functional/config.py
```diff
@@ -91,3 +91,7 @@ class CrossEntropyImpl(str, enum.Enum):
     torch = "torch"
     fused = "fused"
     triton = "triton"
+
+class LossFunctionType(str, enum.Enum):
```
jlamypoirier, line 95: Seems like a bad idea to call it a loss function. DPO doesn't just swap cross-entropy with another function, it changes the argument list altogether. It also affects data sampling and preprocessing.

How about we call it "training objective" instead or something like it?

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -116,13 +160,45 @@ def get(
...
-            sample_spans[:, 0] = np.maximum(sample_spans[:, 0], offset) - offset
+
+            # subtract by offset to normalize span boundaries
+            sample_spans[:, 0] = np.maximum(sample_spans[:, 0], offset) - offset  # offset
             sample_spans[:, 1] = np.minimum(sample_spans[:, 1], offset + len(token_ids) - 1) - offset
-        return GPTSample(token_ids=token_ids, loss_masking_spans=sample_spans)
+
+        chosen_span = None
+        rejected_span = None
+        if use_preference_loss_masking_spans and self._chosen_spans is not None and self._rejected_spans is not None:
```
jlamypoirier, line 175: is there any scenario where `(self._chosen_spans is not None) != (self._rejected_spans is not None)`

fast_llm/data/dataset/gpt/config.py
```diff
@@ -57,6 +57,16 @@ class GPTSamplingConfig(SamplingConfig):
         desc="Read loss masking spans from the dataset.",
         hint=FieldHint.feature,
     )
+    use_preference_loss_masking_spans: bool | None = Field(
```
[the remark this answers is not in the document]
jlamypoirier: Also this might be better suited for `GPTSamplingData`. This way the trainer could set this value based on the training objective and avoid the complicated relationship between the parameters.

fast_llm/data/dataset/gpt/memmap.py
jlamypoirier: Also given the intended usage we really want both to exist when `use_preference_loss_masking_spans=True`, otherwise DPO will crash...

jlamypoirier (commented):
I'm worried too many unrelated things need to be set right for things to work (dataset format, sampling config, loss function), and if not things will crash too late with a cryptic error. Let's try to simplify this a bit.

tobyzl2:
> I'm worried too many unrelated things need to be set right for things to work (dataset format, sampling config, loss function), and if not things will crash too late with a cryptic error. Let's try to simplify this a bit.

Yeah this makes sense.  Let me try to do a bit of refactoring so from the user perspective they will only have to specify one configuration (something like training objective = dpo) and this will automatically set the other flags without having

tobyzl2:
Since we are already saving whether the data has preference pairs within the memmap dataset (https://github.com/ServiceNow/Fast-LLM/blob/dab6dabf59f1b12301b4e0f9dd738830a966b59c/fast_llm/data/dataset/gpt/memmap.py#L46C1-L47C1), I've made a few modifications so that now the user won't have to pass in any extra configurations (use_preference_loss_masking_spans=True, enable_packing=False) and these fields will get populated automatically.  The only config the user needs to specify is which loss function to use and pass in a compatible dataset (will add some checks to ensure this).

fast_llm/data/dataset/gpt/config.py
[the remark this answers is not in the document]
tobyzl2: Just made some changes to remove this extra flag, it's reading it automatically from the flag in the saved memmap dataset now.

fast_llm/data/dataset/gpt/memmap.py
tobyzl2: In terms of expected behavior I think both chosen spans and rejected spans should always be specified.  I've added a few extra checks so that it fails early if either of these are empty when preference spans are enabled.

fast_llm/layers/language_model/head.py
tobyzl2: Yeah I agree, let me see where I can add these checks so that we can detect it earlier.

fast_llm/data/dataset/gpt/config.py
```diff
@@ -57,6 +57,16 @@ class GPTSamplingConfig(SamplingConfig):
...
+        default=None,
+        desc="Read preference loss masking spans from the dataset.",
+        hint=FieldHint.feature,
+    )
+    enable_packing: bool | None = Field(
```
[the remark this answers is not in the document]
tobyzl2: Made some changes to remove this flag, packing would be having multiple documents in the same sequence in pretraining but for DPO wanted to have a way to pad to the end of sequence instead.

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -83,6 +91,36 @@ def _init(self, name: str, prefix: pathlib.Path | str, num_documents: int | None
                     ).reshape(-1, 2)
                 )
 
+        # read preference spans
+        self._chosen_spans = None
+        self._rejected_spans = None
+        if self._has_preference_spans and self._version >= 3:
```
jlamypoirier, line 97: let's just set self._has_preference_spans=False for other versions.

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -116,13 +158,50 @@ def get(
...
-            sample_spans[:, 0] = np.maximum(sample_spans[:, 0], offset) - offset
+
+            # subtract by offset to normalize span boundaries
+            sample_spans[:, 0] = np.maximum(sample_spans[:, 0], offset) - offset  # offset
             sample_spans[:, 1] = np.minimum(sample_spans[:, 1], offset + len(token_ids) - 1) - offset
-        return GPTSample(token_ids=token_ids, loss_masking_spans=sample_spans)
+
+        chosen_span = None
+        rejected_span = None
+        if self._has_preference_spans and self._chosen_spans is not None and self._rejected_spans is not None:
```
jlamypoirier, line 173: Can be simplified to `if self._has_preference_spans` by bringing the elif content here.

fast_llm/data/dataset/gpt/sampled.py
```diff
@@ -73,6 +75,252 @@ def _lazy_load(self):
 TOKEN_CUMSUM_RATE = 10
 
 
+class GPTSampledPreferenceIndexedDataset(SampledDataset):
```
jlamypoirier, line 78: This has most of its code in common with `GPTSampledIndexedDataset, why making a separate class?

fast_llm/models/gpt/model.py
```diff
@@ -286,6 +289,30 @@ def preprocess(
                                     labels[start : end + 1, i] = -100
                                 else:
                                     labels[i, start : end + 1] = -100
+                if batch.chosen_loss_masking_spans is not None:
```
jlamypoirier, line 292: consider moving this to a `Preprocessor` to reduce bloat.

fast_llm/data/data/gpt/data.py
```diff
@@ -154,9 +168,19 @@ def get_iterator(
         Assert.incl(dataset_name, self._datasets)
         Assert.in_range_incl(batch_config.sequence_length, 1, self._max_sequence_length)
         log_main_rank(f"Initializing {dataset_name} dataset iterator from sample {consumed_samples}...")
+
+        dataset = self._datasets[dataset_name]  # noqa
+
+        if hasattr(dataset._dataset, "_indexed_dataset") and hasattr(
```
jlamypoirier, line 174: This won't work, the actual dataset may be wrapped by concatenations, blending, splitting, etc., and there may be multiple base datasets with variable `_has_preference_spans`. Also we don't know if the model even wants spans.

Best option is to add its as an argument to `GPTData.__init__`. Then it can be forwarded to the dataset sampling through `GPTSamplingData`, so they can pick the right sampling method, i.e. skip loading spans if not needed and raise an error if we want spans that aren't there.

jlamypoirier (commented):
That's a lot better, thanks! Some issues remain on the data side.

fast_llm/layers/language_model/config.py
```diff
@@ -140,6 +142,16 @@ class LanguageModelBaseConfig(LanguageModelArchitectureConfig, BaseModelConfig):
         desc="Min value for clamping initialized weights of the vocabulary embedding and output (logits).",
         hint=FieldHint.feature,
     )
+    loss_function: LossFunction = Field(
```
jlamypoirier, line 145: I faced something similar with distillation #229 and in the end decided that a simple enabling flag (variable) was more appropriate. Might be the same here, at least for now.

tscholak:
@tobyzl2 thanks for reviewing #229!
do you see a path for adding a reference model to DPO now? #216 and #229 should enable that.

fast_llm/data/data/gpt/data.py
tobyzl2: Thanks for sharing this @jlamypoirier - just wanted to confirm the general flow with you before implementing the changes

From my understanding we'd pass in a user parameter that would populate `GPTSamplingParameters`, this will help control the logic of whether we are using preference spans and how we're reading memmapped data.

There's also the other aspect where certain training objective require preference loss spans while others might not be compatible.  In this case, we would want to raise an error early when preference loss spans are required but not specified but skip span loading otherwise.

One difference between preference spans and regular loss masking spans is currently if a dataset is written with preference spans, it assumes it is being used since there are both chosen and rejected turns.  We'd have to define what skipping means - do we apply loss mask on the chosen turn or on the entire conversation or should we also just throw an error if a preference dataset is provided to a non-compatible training objective?

fast_llm/layers/language_model/config.py
tobyzl2: By this would you mean a flag to enable dpo loss and by default if this flag is not provided then it would use cross entropy?  In #229 I noticed when distillation model is specified it will automatically toggle TargetFormat type, just checking this is what you were referring to.

jlamypoirier: Yes, or equivalent

fast_llm/data/data/gpt/data.py
jlamypoirier: Ok, raising an error makes sense. 

(The disjoint use case worries me a bit, it looks like an entirely different dataset type rather than a new feature, so separate class(es) would be more appropriate. Llet's ignore for now since the infrastructure isn't ready, but keep it in mind for the future.)

fast_llm/layers/language_model/config.py
tobyzl2: Added a flag `use_dpo_loss` so now it's just a simple toggle between enable/disable with the default being cross entropy loss.

fast_llm/data/data/gpt/data.py
tobyzl2: Yeah makes sense since it's an entirely new training stage/objective.  For now, I've added an early check within config validation to make sure if dpo loss is enabled, preference pairs must be enabled too (and vice versa).  Memmap class also ensures that the dataset actually has these preference pairs or else an error is thrown.

fast_llm/data/dataset/gpt/sampled.py
tobyzl2: I agree it's cleaner all in one class, made the change to consolidate the two.

fast_llm/data/dataset/gpt/memmap.py
```diff
@@ -170,31 +171,34 @@ def get(
 
         chosen_span = None
         rejected_span = None
-        if self._has_preference_spans and self._chosen_spans is not None and self._rejected_spans is not None:
-            chosen_span = self._chosen_spans[idx]
 
-            # filter spans that are outside the range of the selected tokens in the document
-            chosen_span = chosen_span[(chosen_span[0] < offset + len(token_ids)) & (chosen_span[1] >= offset)][0]
+        if use_preference_loss_masking_spans:
+            if self._has_preference_spans and self._chosen_spans is None:
```
jlamypoirier, line 176: can simplify conditions with
```
if not self._has_preference_spans:
    raise ValueError("No preference spans found in memmap dataset.")
```
at the beginning

fast_llm/models/gpt/config.py
```diff
@@ -196,6 +201,9 @@ def _validate(self) -> None:
         if self.model.base_model.distillation_model is not None:
             # TODO: Support loss masking for distillation?
             assert not self.batch.use_loss_masking_spans
+        assert self.model.base_model.use_dpo_loss == self.batch.use_preference_loss_masking_spans
```
jlamypoirier, line 204: Seems to make the two parameters redundant

fast_llm/layers/language_model/head.py
```diff
@@ -146,13 +144,16 @@ def forward(
...
+            target = None
+        else:
+            target = kwargs.get(
+                LanguageModelKwargs.labels
+                if self._use_dpo_loss or self._config.distillation_model is None
+                else f"{self._config.distillation_model}_logits"
+            )
         if target is not None:
-            if self._config.distillation_model is None:
+            if self._config.distillation_model is None or self._use_dpo_loss:
```
jlamypoirier, line 156: Checks for `self._use_dpo_loss` redundant since it doesn't support distillation?

jlamypoirier (approved):

jlamypoirier:
Are we planning on merging this?

tobyzl2:
> Are we planning on merging this?

Yes, seeing if we need to wait to merge #255 first @tscholak ?

tscholak:
Let's merge DPO @tobyzl2.
The other PR is blocked atm

Issue #228, opened 2025-04-10
[bug] Logit scaling incorrect for triton and torch cross-entropy


jlamypoirier:
# 🐞 Describe the Bug

Triton implementation of logit scaling for cross-entropy is incorrect:
* Re-loaded logits are not scaled https://github.com/ServiceNow/Fast-LLM/blame/main/fast_llm/functional/triton/cross_entropy.py#L36
* Scaling gradient is applied at the wrong place (on exp logits instead of final gradient) https://github.com/ServiceNow/Fast-LLM/blame/main/fast_llm/functional/triton/cross_entropy.py#L48

Also the torch implementation crashes with logit scaling or with `grad_output=None`. https://github.com/ServiceNow/Fast-LLM/blob/main/fast_llm/functional/cross_entropy.py#L11.

The bugs have been around since loss scaling was first implemented in e933bd6, so could have impacted experiments.

Found those while adapting cross-entropy for knowledge distillation #214. I already made fixes and tests, I'll include in the PR unless there is an immediate need.

tscholak:
thanks for catching this!
we never used loss scaling, luckily.

PR #229, opened 2025-04-11
Knowledge distillation, fix and improve cross-entropy

jlamypoirier:
# ✨ Description

Fix: #214, #228

* [x] Fix bugs in cross-entropy
* [x] Add tests for cross-entropy
* [x] Extend cross-entropy implementations to support probability distributions instead of logits.
* [x] Implement end-to-end knowledge distillation using a reference model (#216) and a cross-entropy loss between reference and predicted logits.

Basic knowledge distillation with cross-entropy loss.

Known issues/concerns:
* Loss masking is silently ignored because only the labels know about it, and we're ignoring them. (Fixable after #230)
* Preprocessing is shared between the reference model and trained model, and can break if they have different preprocessing options. (Ex. embedding, flash attn, etc.). Added restrictions, but we'll probably want to address, the need for exact match with rotary embeddings (including variables like theta) could be a problem.
* Not too sure about just storing the distillation model in the base model config, and putting the actual model definition in a far away location.
* Enabling distillation disables the standard LM loss. @tscholak @RaymondLi0 is there a scenario where we want both?

## 🔍 Type of change

Select all that apply:


tscholak:
Hi @jlamypoirier,

I realize I didn't give enough context upfront, so I want to clarify the intent behind this task.

I told the team you'd implement one specific knowledge distillation loss, which can show how to wire in the reference model and use its logits in the loss computation. The goal was to create a clear, minimal working example that others (including interns) could follow to implement more loss types later.

What we got instead is a general-purpose loss interface that abstracts over formats, types, and backends. While I see where you're coming from in terms of reusability and performance, this isn't what we need right now, and it actually works against the original goal.

We are missing an explicit, self-contained loss implementation. One that:

* lives in its own file or function,
* computes KL divergence (or similar) between predicted and reference softmax outputs,
* integrates into the training loop without additional abstractions.

Clarity is more important than generality here. The current form hides logic behind config values and makes assumptions that aren't obvious to others. Interns won't be able to extend or adapt it without decoding all the branching behaviour, internal scaling factors, and implicit format conversions.

That's on me for not being more specific at the outset. But now that we've seen where this goes, I want to rewind this PR and reframe it as a clean, minimal first example, not a generalized loss framework.

Let's optimize later. For now, let's land something explicit, readable, and easy to build on.

jlamypoirier:
Knowledge distillation is coming, please wait for it before judging its complexity, there won't be any.

What I've done so far deals with the cross-entropy loss (aka KL divergence) that will go in it. There is nothing fancy about it, it's just cross-entropy, and plugging it in the existing cross-entropy was easy enough. I did however notice that our existing implementation wasn't working properly, so I had to add tests to find out the extent of the problem, and deal with the bug.

fast_llm/engine/training/config.py
```diff
@@ -431,13 +429,12 @@ def _add_reference_distributed_to_pretrained(pretrained: PretrainedFastLLMModelC
...
+        pretrained.model.distributed.validate()
+        Assert.leq(pretrained.model.distributed.to_dict().keys(), {"world_size", "rank", "local_world_size"})
         with NoAutoValidate():
             pretrained.model.distributed = distributed.to_copy()
         # Allow sharing the `Distributed` instance.
         pretrained.model.distributed.reference_config = distributed
         old_setup()
 
-    pretrained._setup = new_setup
+    object.__setattr__(pretrained, "_setup", new_setup)
```
tscholak, line 440: bit of a hack, no?

RaymondLi0:
About MTP + knowledge distillation, one use-case I see would be when doing CPT with more prediction heads:
adding a KL-divergence term would be a way to ensure we do not diverge too much from the pretrained model's outputs by training the new heads. In that case we'd use the lm-loss for all the heads, and the distillation loss on the first head (which also exists in the reference model).
But let's first merge this :)

tscholak:
@jlamypoirier, how does a typical config look like for this?

nitsanluke (approved):
LGTM ❤️ ! 
Minor notes: as a newbie it was bit difficult for me to find the main changes for distillation; if its possible to make the PRs more atomic it would be much easier to review. 
Also having a run (or config) to see the training would be awesome to verify. Thanks!

jlamypoirier:
> About MTP + knowledge distillation, one use-case I see would be when doing CPT with more prediction heads: adding a KL-divergence term would be a way to ensure we do not diverge too much from the pretrained model's outputs by training the new heads. In that case we'd use the lm-loss for all the heads, and the distillation loss on the first head (which also exists in the reference model). But let's first merge this :)

So if I understand correctly the needs would be (for future work):
* Allow mixing LM and distillation losses
* Allow distillation from a model with fewer MTP heads
* KL-divergence loss? It's basically the same as cross-entropy (loss differs by a near-constant tern, grads are the same), would CE work as-is? Or we can adapt the existing CE implementations for KL, it's a 2-3 line change.


> @jlamypoirier, how does a typical config look like for this?

Here is a simple example I ran for testing, using the pretrained model from the quick-start guide. (Note some issues with rotary to be fixed in #235.
```
2025-04-15 00:32:54,111 Command run:
/usr/local/bin/fast-llm train gpt training.train_iters=10000 data.format=random reference_models.teacher.pretrained.path=fast-llm-tutorial/pretrained-model reference_models.teacher.pretrained.format=llama model.base_model.distillation_model=teacher model.base_model.transformer.rotary.type=default model.base_model.cross_entropy_impl=triton
2025-04-15 00:32:54,115 
------- fast_llm.models.gpt.config.GPTTrainerConfig --------
run:
  experiment_dir: null
model:
  base_model:
    transformer:
      normalization:
        type: layer_norm
      peft:
        type: none
      num_layers: 12
      hidden_size: 1024
      num_attention_heads: 8
      head_groups: 1
      add_linear_biases: true
      ffn_hidden_size: 4096
      kv_channels: 128
      rotary:
        type: default
      activation_type: gelu
    vocab_size: 49152
    tie_word_embeddings: true
    cross_entropy_impl: triton
    distillation_model: teacher
  distributed:
    world_size: 1
    rank: 0
    local_world_size: 1
    training_dtype: float32
pretrained:
  format: fast_llm
  path: null
  load_config: architecture
training:
  evaluations: {}
  export:
    format: fast_llm
  train_iters: 10000
batch:
  micro_batch_size: 1
  depth_first_micro_batches: 1
  breadth_first_micro_batches: 1
  sequential_micro_batches: 1
  batch_size: 1
  sequence_length: 2048
data:
  split:
  - 969.0
  - 30.0
  - 1.0
  format: random
  fim:
    rate: 0.0
  tokenizer:
    path: null
  datasets:
    training:
      split:
      - 969.0
      - 30.0
      - 1.0
      format: random
      fim:
        rate: 0.0
      type: legacy
    validation:
      split:
      - 969.0
      - 30.0
      - 1.0
      format: random
      fim:
        rate: 0.0
      type: legacy
    test:
      split:
      - 969.0
      - 30.0
      - 1.0
      format: random
      fim:
        rate: 0.0
      type: legacy
profiling:
  cuda: false
optimizer:
  learning_rate:
    base: 0.0001
  weight_decay: 0.01
reference_models:
  teacher:
    model:
      base_model:
        transformer:
          normalization:
            type: rms_norm
            epsilon: 1.0e-05
          peft:
            type: none
          num_layers: 30
          hidden_size: 576
          num_attention_heads: 9
          head_groups: 3
          add_linear_biases: false
          ffn_hidden_size: 1536
          kv_channels: 64
          rotary:
            type: default
            theta: 100000.0
            scale_factor: 8.0
            low_frequency_factor: 1.0
            high_frequency_factor: 4.0
            original_context_length: 8192
            attention_factor: null
            beta_fast: 32.0
            beta_slow: 1.0
          gated: true
          activation_type: silu
        vocab_size: 49152
        use_position_embeddings: false
        tie_word_embeddings: true
      distributed:
        world_size: 1
        rank: 0
        local_world_size: 1
        training_dtype: float32
    pretrained:
      format: llama
      path: fast-llm-tutorial/pretrained-model
      load_config: architecture
--------------------------- end ----------------------------
```

I'll merge after #211 which is pending approval.

jlamypoirier:
> LGTM ❤️ ! Minor notes: as a newbie it was bit difficult for me to find the main changes for distillation; if its possible to make the PRs more atomic it would be much easier to review. Also having a run (or config) to see the training would be awesome to verify. Thanks!

Somehow the PR got messed up after a merge with main, sorry about that. It's fixed now.

tscholak (approved):
thanks for the contribution and adding the config @jlamypoirier!
LGTM

tobyzl2:
Agreed with the above that combining soft target / LM loss could be a useful feature for the future.  Small question I had was how we are defining `probabilities` vs `logits` in `TargetFormat` - I guess these will automatically be set by presence of distillation model.  Other than that all LGTM!

tobyzl2 (approved):

jlamypoirier:
I'm holding off on merging this one because of an issue with parallel cross-entropy, in the meantime feel free to try out this branch (and #235), it should otherwise work.

jlamypoirier:
Turns out the bug is already in main and was introduced in #218, so I'll go ahead with the merge

PR #230, opened 2025-04-15
Sampling parameters, generalize batch config.

jlamypoirier:
# ✨ Description

Deal with some structural issues and technical debt causing trouble for ongoing work.
* Make a `SamplingParameter` structure for holding the data sampling parameters that come from the trainer (batch config, model, etc.). This is needed as an alternative to the fast-growing argument list to `GPTData.__init__` and the associated bloat. (ex. #177, #218, #223) 
* Generalize `BatchConfig`, extract the model-specific parameters into `GPTBatchConfig`.
* Rename `num_micro_sequences` -> `micro_batch_splits`. Since the generic batch config and schedule runner shouldn't have to know about model-specific sequences.
* Move `use_loss_masking_spans` to the batch config. This will make it easier to know if loss masking is enabled, ex. to prevent it in #229). @sohamparikh this may require some config changes. I added backward compatibility, but it will only work if set globally (`data.sampling.use_loss_masking_spans`)
* Trying out `cached_property` in a few places following the discussion in #211. If it work we can use it all over the place to simplify derived fields.

## 🔍 Type of change

Select all that apply:

### DIFF
--- a/fast_llm/data/data/abstract.py
+++ b/fast_llm/data/data/abstract.py
@@ -4,6 +4,7 @@ import typing
 
 from fast_llm.config import Configurable
 from fast_llm.data.data.config import DataConfig
+from fast_llm.data.dataset.config import SamplingParameters
 from fast_llm.engine.distributed.config import DistributedConfig
 from fast_llm.engine.schedule.config import BatchConfig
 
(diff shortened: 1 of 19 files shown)


tscholak:
I think the most interesting and valuable change here is actually that with this we can now set these so-called sampling parameters differently for different datasets.

tscholak (approved):
LGTM

Issue #237, opened 2025-04-17
Review the architecture split


jlamypoirier:
# 🎯 **Goal (What & Why)**

Base model configs are currently split into architecture/non-architecture. This is a legacy behaviour that was introduced as a stopgap solution for:
1. Loading pretrained configs while allowing changes in parameters that support it.
2. Marking a subset of the parameters as essential to a model, which absolutely need to be kept fixed when loading a pretrained model (i.e. changing it would cause a crash or otherwise break the model).

(1) Is made obsolete by #211, and will become irrelevant after #231. (2) On the other hand needs a replacement. Ignoring the problem is not recommended because it would make the loading of pretrained checkpoints unsafe, causing cryptic crashes and unexpected breakdown. My suggestion (which I've been planning for a while) would be to replace the hard class split by a simple field descriptor, i.e. something like field hints. This would address the issue with relatively low added complexity (and could be made opt-in, ex. for experimental models).

# 🚀 **Execution Plan**
> _(This section may start as an incomplete draft but must be defined before implementation begins.)_ 

### **Step 1: What is the smallest working version?**
> _(Describe the simplest way to implement this feature with minimal effort.)_  

### **Step 2: What additional optimizations are possible (but optional)?**  
> _(List potential refinements that can be added in later PRs if needed.)_  

# 📌 **Acceptance Criteria** (Must-Haves for Completion)
* The feature must be **functional and tested**.  
* The implementation must be **documented in practical terms**.  
* The PR must include a **performance/impact summary**.  
* **No refactors unless directly necessary** for feature completion.  

# 🛠️ **Project Management**
- [ ] **Assign the project to the Fast-LLM project.**
- [ ] **Set the `Estimate` field (in days) in the GitHub project.**
- [ ] **Use the `Size` field to categorize the PR size (Small/Medium/Large).**
- [ ] **Assign an owner when opening the issue.**

PR #245, opened 2025-04-30
[Prototype] Generalize dynamic config classes

jlamypoirier:
# ✨ Description

Fix: #126

Generalize the concept of dynamic config class from the dataset mechanism to all config class. I opted for a unique global registry of all config classes. This choice is motivated by simplicity (multiple registries would be hell to manage), though it means registry keys need to be globally unique so they'll tend to be longer. Config classes are now specified through the `type` field, using the config class name or its shorthand without the `Config` suffix.

Included a proposal to simplify the cli, so the runnable is selected through its `type` field, ex. `fast-llm type=GPTTrainer ...` instead of `fast-llm train gpt`.

Added backward compatibility for the old dataset dynamic type names.

TODO:
*  Add backward compatibility for the cli?
* The `type` field conflicts with normalization, rotary and peft types. These are good use cases for dynamic classes so I'll implement, but I'll wait after #237.
* Some other potential use cases in the optimizer (Lr schedule), but no need to do in this PR.

## 🔍 Type of change

Select all that apply:


jlamypoirier:
@tscholak I started working on more dynamic classes and realized user-friendly names are essential. So I implemented a simple solution where each class can have its own registry, and the class is selected by iterating through the registries in the MRO. We still build a `Config` registry from class names, and others can be created with a simple `register` call.

So we get neat names without adding complexity on the developer side. The only real missing piece would some way for users to find out the available types, we should add it to automatic config documentation once we have it.

tscholak:
yes, this is great!

fast_llm/cli.py
```diff
@@ -0,0 +1,48 @@
...
+    # TODO: Add hook to register model classes? (environment variable?)
+    # (Pre-)configure logging
+    configure_logging()
+    try:
+        if args is None:
+            args = sys.argv[1:]
+        # TODO: Remove backward compatibility.
+        if len(args) >= 2 and args[0] == "train":
+            if args[1] == "gpt":
+                args = ["type=train_gpt"] + args[2:]
```
tscholak, line 28: with regards to deprecating the `fast-llm train` command syntax. I'm not super confident that we ever will.

fast_llm/data/dataset/gpt/config.py
```diff
@@ -558,3 +494,18 @@ def build_and_sample(self, sampling: SamplingData) -> SampledDataset:
...
+GPTSampledDatasetConfig.register_subclass("memmap", GPTMemmapDatasetConfig)
+GPTSampledDatasetConfig.register_subclass("concatenated", GPTConcatenatedDatasetConfig)
+GPTSampledDatasetConfig.register_subclass("slice", GPTDatasetSliceConfig)
+GPTSampledDatasetConfig.register_subclass("sampled", GPTSampledDatasetUpdateConfig)
+GPTSampledDatasetConfig.register_subclass("blended", GPTBlendedDatasetConfig)
+GPTSampledDatasetConfig.register_subclass("file", GPTDatasetFromFileConfig)
+GPTSampledDatasetConfig.register_subclass("concatenated_memmap", GPTConcatenatedMemmapConfig)
+GPTSampledDatasetConfig.register_subclass("fim", GPTFimSampledDatasetConfig)
+GPTSampledDatasetConfig.register_subclass("legacy", GPTLegacyDatasetConfig)
+GPTSampledDatasetConfig.register_subclass("test_slow", GPTTestSlowDatasetConfig)
```
tscholak, line 511: hmm, so this is for registering user-friendly names or the classes themselves?
can we make registration automatic? didn't we want to use the `type` classvar for that?

fast_llm/engine/training/config.py
```diff
@@ -349,30 +338,23 @@ class TrainerConfig(PretrainedFastLLMModelConfig, ExperimentConfig):
     _abstract = True
     # TODO: Generalize data, schedule, logging, etc.
     training: TrainingConfig = Field(
-        default_factory=TrainingConfig,
```
tscholak, line 352: you mentioned in the other PR that this is necessary now, but I don't quite understand why. if you've got several classes of type `TrainingConfig`, you want to have a straightforward way to say which one is the default. that's pretty much it

fast_llm/layers/transformer/config.py
```diff
@@ -114,48 +132,153 @@ class RotaryConfig(BaseModelConfig):
...
+    Yarn scaling:
+    https://github.com/huggingface/transformers/blob/006d9249ec0270ff6c4d3840979d23fe94bdc763/src/transformers/modeling_rope_utils.py#L163
+    [original paper](https://arxiv.org/abs/2309.00071)
+    """
+
+    # TODO: Add descriptions.
+    scale_factor: float = Field(default=8.0, hint=FieldHint.feature)
     attention_factor: None | float = Field(
         default=None,
-        desc="Attention factor for yarn-type scaling.",
```
tscholak, line 132: is this because the desc is not informative or because something has changed so that descriptions don't work anymore?

fast_llm/layers/transformer/config.py
```diff
@@ -177,8 +300,49 @@ class TransformerSubLayerName(str, enum.Enum):
...
+
+    def apply_weight(self, parameter: "ParameterMeta") -> "ParameterMeta":
+        return parameter
+
+
+@config_class()
+class TransformerLoRAConfig(LoRAConfig, TransformerPeftConfig):
     layers: list[TransformerSubLayerName] = Field(
-        default=None,
+        default=(TransformerSubLayerName.query, TransformerSubLayerName.value_),
```
tscholak, line 345:
```suggestion
        default_factory=lambda: [TransformerSubLayerName.query, TransformerSubLayerName.value_],
```

jlamypoirier: The tuple is fine, it gets transformed into a list automatically

fast_llm/layers/transformer/config.py
jlamypoirier: `Attention factor` is just repeating the field name, and `for yarn-type scaling` is now redundant with the type/class name. So the existing descriptions are completely useless.

fast_llm/cli.py
jlamypoirier: Found a way to preserve the syntax https://github.com/ServiceNow/Fast-LLM/pull/245/files#diff-1572906edaf7e94bf94dcf9dfdecc5b34757c2e56924cd3054a365b10db5ee31R26

fast_llm/data/dataset/gpt/config.py
jlamypoirier: `type` is for the instance field. We could add other classvars, but it wouldn't be ideal because we'd still need to distinguish which base class it belongs to (ex. one classvar for each base class), and we may want to register multiple user-friendly names. Also It adds the magic of the classvar being converted to dynamic type name. Overall the external register call seems better.

tscholak: I am confused by this answer. I was expecting something like this:

If a config class defines/overrides the `type` classvar field, it automatically becomes a registered dynamic class with the value of the `type` field becoming its human-readable name, which will be used for serialization and deserialization.

Example:

```python
@config_class
class MainConfig:
    foo: FooConfig

@config_class
class FooConfig:
    type: typing.ClassVar[str]

class BarConfig(FooConfig)
    type = "bar"
    something: int

class BazConfig(FooConfig)
    type = "baz"
    something_else: int
```

Serialized version:

```yaml
foo:
  type: baz
  something_else: 5
```

fast_llm/layers/transformer/config.py
tscholak: ok, redundancy, I get it. But rather than dropping the description entirely, should we consider replacing it with something more informative? e.g., when this field matters, what values are valid, what happens if it's None, etc. Otherwise we're losing a chance to document intent.

fast_llm/layers/transformer/config.py
tscholak: I'm OK with the tuple **if** there's a documented guarantee that the config layer converts it automatically and that states the conversion rules clearly.

Right now, the behaviour isn't obvious at all. The type of this field is annotated as a list but initialized as a tuple. That's a straightforward type mismatch. I wasn't aware that coercion rules existed beyond serialized data, and I doubt anyone else is either. Does it always convert tuple to list? What happens with `typing.Sequence` or other container types?

Explicit is always better than implicit. Let's document these conversions clearly, or just avoid them entirely by using explicit defaults.

fast_llm/data/dataset/gpt/config.py
tscholak: I think unless we can show a concrete need for multiple human-readable names per class or for registering the same subclass under multiple base classes (imho both extremely niche), the explicit `register_subclass(...)` pattern adds too much noise and duplication.

fast_llm/layers/transformer/config.py
jlamypoirier: Yes, this will be part of the config documentation.

fast_llm/data/dataset/gpt/config.py
jlamypoirier: `type` is a field for instances https://github.com/ServiceNow/Fast-LLM/pull/245/files#diff-15aab2503b9a097809b7fa450b3f63ea0db5d700982d7265610fdf28b476788aR989. What you are referring to is an equivalent to the `type_` classvar we had before https://github.com/ServiceNow/Fast-LLM/pull/245/files#diff-94ac1cad97ec073f07a7d6ead2489610430a6fb40aa3e34d520b86bd88e6c9fcL102. These two can't be the same. 

Adding back the same `type_` variable is no longer an option, because all base classes have a registry and we don't know which one is concerned. So we'd need some adjustments, I can think of a few options: a `registry_class` classvar, separate type variables (ex. `gpt_dataset_type`), or a more complex `type: typing.ClassVar[tuple[tuple[type[Config], str], ...]]`. None of these look better than `register_subclass`.

I think we do want to support multiple names and/or base classes. Multiple names may be needed for backward compatibility, or to simplify usage. (The final saved config use the full class name either way, so that's just convenience.) And multiple base classes is useful for specialization, for example `GPTTrainerConfig` should be registered as `train_gpt` as a `RunnableConfig`, but can take the shorter name `gpt` as a `TrainerConfig`. Concretely this allows the `fast-llm train gpt` syntax.

tscholak: I think we are heading toward unnecessary complexity that harms the developer experience of the config system.

It's pretty clear that the root issue is that we're treating **every** config class as a dynamic (tagged union) config, when in reality 99% of them have a single concrete subclass and don't need any dynamic dispatch. This design choice forces every base class to maintain a registry (even when it doesn't need one), and it forces every developer to think about tagging, naming, registration, and base-class implications just to define a simple config. That's complexity overhead we shouldn't be paying.

If dynamic dispatch is needed (and I agree that it sometimes is), it should be opt-in via a mixin or decorator, e.g.:

```python
class DynamicConfigMixin: ...  # adds registry + tag handling

class FooConfig(BaseConfig):   # plain, no registry
    ...

class BarConfig(DynamicConfigMixin, BaseConfig):  # dynamic
    tag: ClassVar[str] = "bar"
```

I suggested this earlier. That gives us control: only configs that actually benefit from dynamic subclassing pay the complexity cost. Everyone else just writes a normal, simple config class code.

This would also clean up the type/registry ambiguity, avoid the diamond/multiple-inheritance traps, and decouple config structure from CLI syntax or UX naming (which should live elsewhere anyway).

You mentioned cross-registry cases. Yes, they can be handled by an explicit registries ClassVar. It's not pretty, but it works. And this is fine, because that should be well under 1% of the use cases (if it isn't, we have a bigger problem), and this edge case shouldn't dictate the default.

In short: Let's not optimize the config system for the most complex scenario. Let's make the common case simple.

tscholak: Concretely, the example for "one class / two registries" was:

> `GPTTrainerConfig` should be registered as `train_gpt` as a `RunnableConfig`, but can take the shorter name `gpt` as a `TrainerConfig`. This would allow the `fast-llm train gpt` syntax.

So the rationale is:

* One config class (`GPTTrainerConfig`) implements two roles: it's both a `RunnableConfig` and a `TrainerConfig`.
* It should be registered under two different tags, in two different registries, so the CLI can accept `train gpt` and route correctly.

I think type hierarchy should not be shaped by CLI syntax ergonomics. We should separate the two cleanly. So let's keep a separate registry for `fast-llm` commands. If we don't, then the UX demands leak into the whole configuration approach and add complexity that needs to be paid throughout the whole codebase.

jlamypoirier: > I think we are heading toward unnecessary complexity that harms the developer experience of the config system.
> 
> It's pretty clear that the root issue is that we're treating **every** config class as a dynamic (tagged union) config, when in reality 99% of them have a single concrete implementation and don't need any dynamic dispatch. This design choice forces every base class to maintain a registry (even when it doesn't need one), and it forces every developer to think about tagging, naming, registration, and base-class implications just to define a simple config. That's complexity overhead we shouldn't be paying.
> 

This doesn't make much sense to me. My implementation adds exactly zero complexity for the developer when creating a config, there is nothing to do in any situation. All there is is an optional registration after the class definition, which is standard and common practice. 

It's not necessary to give a registry to every config, I made the choice precisely so the developer doesn't have to worry about it (same with most design choices here).

> If dynamic dispatch is needed (and I agree that it sometimes is), it should be opt-in via a mixin or decorator, e.g.:
> 
> ```python
> class DynamicConfigMixin: ...  # adds registry + tag handling
> 
> class FooConfig(BaseConfig):   # plain, no registry
>     ...
> 
> class BarConfig(DynamicConfigMixin, BaseConfig):  # dynamic
>     tag: ClassVar[str] = "bar"
> ```
> 
> I suggested this earlier. That gives us control: only configs that actually benefit from dynamic subclassing pay the complexity cost. Everyone else just writes a normal, simple config class code.

To me this is both less powerful and more complex than my implementation. The developer has to worry about the mixin, multiple inheritance, tagging, magic happening behind the scene, etc. That's a lot more than a register call.

> You mentioned cross-registry cases. Yes, they can be handled by an explicit registries ClassVar. It's not pretty, but it works. And this is fine, because that should be well under 1% of the use cases (if it isn't, we have a bigger problem), and this edge case shouldn't dictate the default.
> 

I actually expect them to be common. Specialization is all over the place in our codebase, and we'll definitely have overlapping registries.

fast_llm/layers/transformer/config.py
jlamypoirier: These descriptions were already irrelevant, so whether I'm removing them or leaving them as a repetition of the field name doesn't matter. And more importantly, I'm not familiar enough with these classes to make a proper description, feel free to suggest some.

fast_llm/data/dataset/gpt/config.py
nandahkrishna: Chiming in here because dynamic configs would be useful for some of the hybrid model architectures we're trying to add support for.

I feel like the explicit `register_subclass` calls might add additional maintenance burden: you would have to remember to add the call to this method and it would be separate from the definition of the config class itself. On the other hand, using a mixin to explicitly indicate the requirement for dynamic config support, and keeping human-readable names within a ClassVar just seems way more simple to me as a developer adding new config classes and architectures to Fast-LLM.

As for multiple human-readable names under different base classes, I still think having a ClassVar that defines this mapping would be easier, and we could keep things concise and within the class definition. I tend to agree with Torsten that allowing multiple human-readable names and as a subclass of several different classes seems like we're offering too much flexibility and thus increasing complexity.

It is possible I'm missing other nuances here so do let me know if I'm missing something.

fast_llm/engine/training/config.py
jlamypoirier: Problem was all sorts of potential issues happening if not going through `_from_dict`, not necessarily related to this PR (ex. dataclasses doing the wrong thing when calling it for nested configs). But now that I enforced going through it (with the metaclass `__call__`), I think I can safely bring it back.

jlamypoirier: Actually this won't work without some hacking around. Problem is we don't just want to set a default, we want to set a default to the nested config's `type` field even is a value is provided. And we can't just set a default to `type` because it would propagate to subclasses and cause all sorts of trouble.

tscholak: Ok, though we still need a clean, consistent way to specify default values for config fields.

```python
field: SomeConfig = Field(default_factory=SomeConcreteConfig)
```
should just work. If `SomeConcreteConfig` is a subclass in dynamic dispatch / tagged union style setup, then instantiating it directly should still produce a valid config object. The fact that it has a type or tag field shouldn't block the default from being usable. If it's not currently reliable, that's a sign that we're over-complicating the dispatch mechanism than a reason to give up on defaults.

jlamypoirier: This would work perfectly fine, just not in a useful way, and it would be a ton of trouble.

`default_factory` (and `default`) is for when a value is *not provided*. This means it is (and should be) ignored whenever anything is passed, including an empty dict. So it's not suitable for setting the default field type, because setting any value will undo it. 

On top of that, the config mechanism doesn't distinguish whether nested fields are passed explicitly or just empty. So some defaults may lead to incorrect serialization paths, ex: `{} --(instantiate)-> Config({}) --(serialize)->  [MISSING] --(instantiate)->  Config([DEFAULT])`. (in fact this happens for all nested fields, not just configs, but configs are the biggest risk)

jlamypoirier: Also note that this doesn't really have to do with dynamic classes or the present PR. It's an instantiation thing, and setting non-dynamic defaults would lead to the same problem, ex: `[Field(default_factory=lambda: SomeConcreteConfig](field: SomeConfig = Field(default_factory=lambda : SomeConcreteConfig(some_field=3))` would have the same problem. This wasn't working before either, all I did here is to replace the silent errors with explicit checks.

tscholak: Hm, the issue you're describing is surfacing because of how dispatch and serialization are currently wired. Let's look at the example you gave:

```
{} --(instantiate)-> Config({}) --(serialize)-> [MISSING] --(instantiate)-> Config([DEFAULT])
```

This only becomes a problem because the **serializer aggressively drops fields** when their values match defaults. In this case, it drops the *entire field*, including the tag, just because the config looks "default-ish." That's the actual cause of round-trip inconsistencies, and not the use of `default_factory`!

We should not be optimizing serialized configs for brevity at the cost of correctness or developer sanity. I've raised this before, many times actually, and I'll say it again: **nobody cares about shaving lines off a YAML file**, especially if it leads to complexity and development headaches.

What we need instead is what every normal config system does:

* If the field is missing or explicitly set to `{}`, call `default_factory()` and use the result.
* If a dict is provided without a tag, use `default_factory()` and apply the provided keys as overrides.
* If a dict includes a tag, use that to resolve the correct subclass and instantiate it with the overrides as usual.

This works the same for dynamic and non-dynamic configs. It produces clean, predictable defaults, supports partial overrides, and round-trips cleanly — **as long as we stop deleting structure during serialization**.

If our config system can't support this behavior without hacks, that's a design flaw in the dispatch/serialization layer and not a reason to abandon reasonable defaulting behaviour.

fast_llm/data/dataset/gpt/config.py
oleksost: Alternatively, to keep subclass registration closer to class definition, we could consider using an additional decorator for this? e.g. something simple like `@register_subclass("fim")`, which would register the subclass under name `fim` in all its base classes that have a registry.

fast_llm/engine/training/config.py
jlamypoirier: Let's not get sidetracked with what we serialize. As I mentioned multiple times already, **I do care a lot about it**, clear and concise configs are essential to me for development, testing, debugging, understanding bug reports, etc. There are plenty of options to get the whole config when you want it. And as for the issue mentioned here, they are easily fixable, but it's not urgent.


> What we need instead is what every normal config system does:
> 
> * If the field is missing or explicitly set to `{}`, call `default_factory()` and use the result.
> * If a dict is provided without a tag, use `default_factory()` and apply the provided keys as overrides.
> * If a dict includes a tag, use that to resolve the correct subclass and instantiate it with the overrides as usual.

Afaik that's **not an established standard**.There are two perfectly good ways to see defaults for config:
1. Config as arguments: Defaults are used when no value is provided. This is the standard for dataclasses, pydantic `BaseModel`, and coding in general.
2. Config as update: The default determines a baseline, which the config updates in a nested way. This is the standard for omegaconf/hydra.

As of now, Fast-LLM lies deeply in option 1. Changing this is not impossible but would be complex and have deep implications, and the discussion on this lies far outside the scope of this PR. 

Either way, switching to option 2 would not help with the present problem of selecting a default subclass, in fact it would make things worse. Because once we pick and create a subclass (say B) for a field (of type A), we can't swap with any subclass C of A, because we committed to defining the fields of class B, i.e. we C must be a subclass of B. (This is what omegaconf does.) So we're not setting a default subclass but rather overriding the type hint, so we might as well just set the type hint right to begin with. In option 1 we don't have this problem because we use the actual type from the beginning.

What we actually want is a default for the `type` field, ant it's just like any field default but needs a slightly different handling since it needs to be set prior to instantiation.

tscholak: I've spent many hours this week diving into this PR and its discussions to understand the broader impact. I want to be crystal-clear about where we stand:

**Fast-LLM is an model training framework, not a configuration R&D project.**

Yet the config subsystem keeps burning days of senior time. That's time we're not spending on training models, fixing bugs, or shipping features. We've now burned more than a week on a problem Pydantic solves out of the box. That's not sustainable.

Every extra layer of custom logic raises the bar for contributors, concentrates knowledge in one person, and stalls the real work: training models and shipping features. As the team lead, I have to call it what it is: it's an *operational risk.*

We need to reverse this pattern. The config system should get out of our way, not pull the whole team into its gravity well.

And no, this isn't about shutting down ideas. It's about protecting this team from complexity creep that slows everyone down.

> As I mentioned multiple times already, I do care a lot about it, clear and concise configs are essential to me for development, testing, debugging, understanding bug reports, etc.

This illustrates the underlying issue: the current design optimizes for what the framework maintainer finds convenient for himself, but not for what contributors actually need when developing models or features.

Now, let's look at what works and doesn't work in Pydantic:

```python
from pydantic import BaseModel, Field, ValidationError
from typing import Literal, Union

class Foo(BaseModel):
    type: Literal['foo'] = 'foo'
    x: int = 1

class Bar(BaseModel):
    type: Literal['bar'] = 'bar'
    y: int = 2

Item = Union[Foo, Bar]

class Wrapper(BaseModel):
    item: Item = Field(default_factory=Foo, discriminator='type')

print(Wrapper().model_dump())  # works, default_factory runs, uses Foo()
# print(Wrapper(item={}))  # doesn't work, validation fails because of missing discriminator
# print(Wrapper(item={'x': 42}))  # doesn't work, validation fails because of missing discriminator
print(Wrapper(item={'type': 'bar', 'y': 99}).model_dump())  # works, dispatches to Bar
```

This setup gets most things right:

* Dynamic dispatch is opt-in.
* The discriminator is explicit and user-controllable.
* Defaults work when the field is absent.
* There is no magic registry.

Does it handle every edge case? No. But it works well for the common case. And that's the bar Fast-LLM needs to meet.

Right now, we're debating edge cases that haven't materialized (e.g., sub-subclassing edge-cases should be rare, and if we're running into them, that's a sign to simplify the class hierarchy and not to complicate things further.) Meanwhile, the team is stuck. That's the wrong tradeoff.

In hindsight, we should have adopted Pydantic from the start. But since we're here now, and deadlines are real, we need to course-correct pragmatically.

We've seen the cost of keeping this overly flexible. It's time to simplify. Here are the principles we need to lock in now:

1. Dynamic config behaviour must be opt-in, via a mixin or decorator. Most config classes will just be plain config dataclasses.

2. Subclasses declare their own tag via something, a `ClassVar[str]`, or decorator param. It doesn't really matter. Your choice, @jlamypoirier. But it must be self-contained and discoverable.

3. If a config field is missing, we use `default_factory`. If a dict is provided without a tag (it shouldn't), we either:
    * inject the default tag automatically (preferred), or
    * error clearly with a message explaining the missing tag (acceptable fallback).

4. No global subclass registries tied to CLI routing. CLI dispatch should be its own, explicit layer. Config structure should not bend to command-line ergonomics.

These constraints exist to make Fast-LLM maintainable and usable by the whole team. The bar is not "can @jlamypoirier make this work." It's "can any contributor pick this up, understand it, and extend it easily."

That's what makes this framework sustainable.

jlamypoirier: Let's be clear, the time we're taking is due to the overly detailed discussion we're having here. I had a working solution 2 weeks ago, with near-zero complexity on the user side. 

> We've seen the cost of keeping this overly flexible. It's time to simplify. Here are the principles we need to lock in now:
> 
> 1. Dynamic config behaviour must be opt-in, via a mixin or decorator. Most config classes will just be plain config dataclasses.

It's already opt-in, there is nothing to do when we don't want it, and nothing will happen if we don't use it. Everything else is implementation details. 

> 2. Subclasses declare their own tag via something, a `ClassVar[str]`, or decorator param. It doesn't really matter. Your choice, @jlamypoirier. But it must be self-contained and discoverable.

Already done, see the decorator implementation.

> 3. If a config field is missing, we use `default_factory`. If a dict is provided without a tag (it shouldn't), we either:
>    
>    * inject the default tag automatically (preferred), or
>    * error clearly with a message explaining the missing tag (acceptable fallback).

That's already what we're doing.

> 4. No global subclass registries tied to CLI routing. CLI dispatch should be its own, explicit layer. Config structure should not bend to command-line ergonomics.

It is and always had been the opposite, the cli adapts to the config system. I updated the cli system according to what the config sytstem now supports, and your request to preserve the syntax

tscholak: Thanks, @jlamypoirier. I want to respond clearly to your points:

> "I had a working solution 2 weeks ago..."

The PR is still in draft, tests were failing non-stop until the last commit, and key review feedback -- including on default handling, tagging, and registration -- remains unresolved or blatantly ignored. There was no point at which this was merge-ready. It stalled until I pushed.

> "There is nothing to do when we don't want [dynamic dispatch]..."

That's not what "opt-in" means. Right now, every config class implicitly enters a world with base-class registries, tag handling, and dynamic behaviours -- even if it doesn't use them. Every single config class pays this complexity tax. That's *passive inclusion*, not opt-in.

What I've asked for is explicit control: if a developer doesn't add a mixin or decorator, the config class behaves like a plain (config) dataclass. No registry, no tagging, no indirection. That's what "opt-in" means in a maintainable system. And I had described how this can work eons ago: https://github.com/ServiceNow/Fast-LLM/pull/104#issuecomment-2599496346

> "Already done..." / "That's already what we're doing."

If that were the case, this conversation wouldn't be happening. Are the default factories back? No? Well then.

> “The CLI adapts to the config system..."

We shouldn't even be debating this. Why is there even coupling between the CLI commands and the config system? We have 3 or 4 commands right now. Give them one, hard-coded root config class each and call it a day.

We're deep in this because we didn't push for simplicity early. I'm asking for a course correction now, not to block you, but to unblock everyone else.

jlamypoirier: > That's not what "opt-in" means. Right now, every config class implicitly enters a world with base-class registries, tag handling, and dynamic behaviours -- even if it doesn't use them. Every single config class pays this complexity tax. That's _passive inclusion_, not opt-in.
> 
> What I've asked for is explicit control: if a developer doesn't add a mixin or decorator, the config class behaves like a plain (config) dataclass. No registry, no tagging, no indirection. That's what "opt-in" means in a maintainable system. And I had described how this can work eons ago: [#104 (comment)](https://github.com/ServiceNow/Fast-LLM/pull/104#issuecomment-2599496346)

Let's not debate about semantics. Nothing visible happens if the developer doesn't do anything, what happens in the background is irrelevant. Unless I'm missing something?

> If that were the case, this conversation wouldn't be happening. Are the default factories back? No? Well then.

Default factories were never there for config classes. All I can do at this time is force them to be an exact match with the type hint, which is equivalent to them not being there. Anything more is not an option.

> We shouldn't even be debating this. Why is there even coupling between the CLI commands and the config system? We have 3 or 4 commands right now. Give them one, hard-coded root config class each and call it a day.

I don't get what's wrong here. I can revert the changes if you want. But we have a perfect match between cli commands and the new dynamic config system, it would be weird to keep and maintain a separate infrastructure just because we want to.

tscholak: Thanks for the update, @jlamypoirier. I don't want to drag this out further, so we'll move forward for now.

That said, I want to clearly record the unresolved concerns:

Every config class now implicitly participates in dynamic dispatch and registry logic, even when that complexity isn't needed. The cognitive overhead of the config system remains too high, and this PR made it worse. The sheer review burden here shows how far the design diverges from what most developers expect in a config system. We've burned over a week of senior time debating mechanics, and that time we could have been spent on models, training, or features.

We continue to build and maintain a bespoke config layer, which distracts from model development. A better choice would be to lean on mature solutions like Pydantic, which are battle-tested and solve 95% of our use cases with far less friction.

If the complexity tax of the current system continues to stall progress, I will propose a full migration to Hydra + OmegaConf + Pydantic. 

For now, I'm prioritizing unblocking the team and moving forward.

fast_llm/data/dataset/gpt/config.py
bigximik: Having a decorator looks very good. However, why can't we use one global registry for all `Config` classes?  
Names should be unique per class hierarchy, conflicts will rise on register (not implemented in example below).

Below is an example using a global `ConfigRegistry`, which supports:
- Command runners
- Typed class hierarchies (models, trainers, evaluators, etc.)
- Instantiating proper subclasses from YAML
- Auto-dispatching based on `type:` keys in config

The system includes demo configs and commands at the end to show how it works in practice.

### Features

- Decorator for easy class registration: `@ConfigRegistry.register("name", implementation_class)`
- Automatic resolution of subclassed dataclasses from YAML with `type` field
- Reusable base logic for any component: models, runners, evaluators, checkpoints, etc.
- Command line integration with flexible structure

### 🔧 Subclass Resolution Logic

In this design, the **"base class"** used to resolve registered subclasses is determined by context:

- ✅ In nested dataclass fields (e.g., `model: ModelConfig`), the base class is what's declared in the field's type annotation.

- ✅ In top-level commands or manual use, the base class is explicitly passed to `from_dict(base_cls, data)` or `ConfigRegistry.get(type_name, base_cls)`.

This allows the same global registry to work across unrelated class hierarchies (e.g., `ModelConfig`, `TrainerConfig`, `EvaluatorConfig`), because each lookup is scoped to its specific base class.


What do you think?

```python
from dataclasses import is_dataclass, fields, dataclass
from typing import get_args, get_origin, Union, Any
import yaml

# ---------------- CONFIG REGISTRY MACHINERY (SIMPLIFIED) ----------------

@dataclass
class Config:
    # Each Config can optionally define a "configurable" runtime class it instantiates
    configurable_class = None

    def get_configurable(self):
        # Returns an instance of the associated runtime class using self as config
        if self.configurable_class is None:
            return None
        return self.configurable_class(self)


class ConfigRegistry:
    # Global registry mapping friendly names (like "gpt", "train") to config classes
    _table: dict[str, set[type]] = {}

    @classmethod
    def register(cls, name: str | None = None, configurable_class: type | None = None):
        # Decorator to register a Config class under a name and link to its runtime class
        def deco(target: Config):
            key = name or target.__name__
            target.configurable_class = configurable_class
            cls._table.setdefault(key, set()).add(target)
            return target
        return deco

    @classmethod
    def get(cls, name: str, base: type) -> type:
        # Retrieve the config class registered under `name` that is a subclass of `base`
        for cand in cls._table.get(name, ()):
            if issubclass(cand, base):
                return cand
        raise KeyError(f"No registered class named '{name}' that is subclass of {base.__name__}")


def from_dict(cls: type, raw: dict | list | Any) -> Any:
    """Recursively convert a dict (or list) into a dataclass instance."""
    if is_dataclass(cls) and isinstance(raw, dict):
        init_kwargs = {}
        for f in fields(cls):
            field_value = raw.get(f.name)
            field_type = f.type

            if field_value is None:
                init_kwargs[f.name] = None
                continue

            init_kwargs[f.name] = _convert_value(field_value, field_type)

        return cls(**init_kwargs)

    return _convert_value(raw, cls)


def _convert_value(value: Any, target_type: type) -> Any:
    # Handles typing annotations like list[T], dict[K, V], Union, and dataclasses
    origin = get_origin(target_type)
    args = get_args(target_type)

    if origin is list and isinstance(value, list):
        return [_convert_value(v, args[0]) for v in value]

    elif origin is dict and isinstance(value, dict):
        return {_convert_value(k, args[0]): _convert_value(v, args[1]) for k, v in value.items()}

    elif is_dataclass(target_type) and isinstance(value, dict):
        # If a 'type' key is present, use it to find a registered subclass
        typ_name = value.pop("type", None)
        if typ_name:
            subclass = ConfigRegistry.get(typ_name, target_type)
            if subclass:
                return from_dict(subclass, value)
        return from_dict(target_type, value)

    elif origin is Union:
        # Handle Optional[T] (which is Union[T, None])
        non_none_args = [arg for arg in args if arg is not type(None)]
        if len(non_none_args) == 1:
            return _convert_value(value, non_none_args[0])

    return value


# ---------------- MODEL CLASSES ----------------

@dataclass
class CheckPointConfig(Config):
    path: str


@dataclass
class GPTCheckPointConfig(CheckPointConfig):
    pass


@dataclass
class SSMCheckPointConfig(CheckPointConfig):
    pass


@ConfigRegistry.register("llama")
@dataclass
class LlamaCheckPointConfig(GPTCheckPointConfig):
    pass


@ConfigRegistry.register("mamba")
@dataclass
class MambCheckPointConfig(SSMCheckPointConfig):
    pass


# Model runtime classes (used during execution)
class Model: pass
class GPTModel(Model): pass
class SSMModel(Model): pass


@dataclass
class ModelConfig(Config):
    checkpoint: CheckPointConfig | None = None

    def get_model(self):
        return self.get_configurable()


@ConfigRegistry.register("gpt", GPTModel)
@dataclass
class GPTModelConfig(ModelConfig):
    pass


@ConfigRegistry.register("ssm", SSMModel)
@dataclass
class SSMModelConfig(ModelConfig):
    pass


# ---------------- TRAINER CLASSES ----------------

@dataclass
class TrainerConfig:
    train_iters: int


# ---------------- EVALUATOR CLASSES ----------------

class Evaluator: pass
class EvaluatorLoss(Evaluator): pass
class EvaluatorLmEval(Evaluator): pass


@dataclass
class EvaluatorConfig(Config):
    def get_evaluator(self):
        return self.get_configurable()


@ConfigRegistry.register("loss", EvaluatorLoss)
@dataclass
class EvaluatorLossConfig(EvaluatorConfig):
    pass


@ConfigRegistry.register("lm_eval", EvaluatorLmEval)
@dataclass
class EvaluatorLmEvalConfig(EvaluatorConfig):
    pass


# ---------------- RUNNER CLASSES ----------------

class Runner:
    def __init__(self, cfg):
        self.cfg = cfg

    def run(self):
        raise NotImplementedError


class EvaluateRunner(Runner):
    def run(self):
        print("Running evaluation for", self.cfg.model.__class__.__name__)
        print("Checkpoint is", self.cfg.model.checkpoint.__class__.__name__)
        for name, ev in self.cfg.evaluations.items():
            print(f" • {name}: {ev.__class__.__name__}")


class TrainRunner(Runner):
    def run(self):
        print("Running Training for", self.cfg.model.__class__.__name__)
        print("Checkpoint is", self.cfg.model.checkpoint.__class__.__name__)
        print("Iterations:", self.cfg.trainer.train_iters)


@dataclass
class RunnerConfig(Config):
    def get_runner(self):
        return self.get_configurable()


@ConfigRegistry.register("evaluate", EvaluateRunner)
@dataclass
class EvaluateConfig(RunnerConfig):
    model: ModelConfig
    evaluations: dict[str, EvaluatorConfig]


@ConfigRegistry.register("train", TrainRunner)
@dataclass
class TrainConfig(RunnerConfig):
    model: ModelConfig
    trainer: TrainerConfig


# ---------------- CLI ENTRYPOINT ----------------

class Command:
    @staticmethod
    def run(argv: list[str]):
        if len(argv) != 2:
            raise SystemExit("Usage: <command> <yaml-string>")

        cmd_name, yaml_blob = argv
        raw_cfg = yaml.safe_load(yaml_blob)

        # Look up the appropriate config class from registry
        cmd_cfg_cls = ConfigRegistry.get(cmd_name, RunnerConfig)

        # Parse YAML to a typed dataclass
        cmd_cfg = from_dict(cmd_cfg_cls, raw_cfg)

        # Instantiate and run the corresponding Runner
        cmd_cfg.get_configurable().run()


# ---------------- DEMO ----------------

evaluate_command_args = """
model:
    type: gpt
    checkpoint:
        type: llama3
        path: /path/to/model
        
evaluations:
    stack3b:
        type: lm_eval
    fineweb:
        type: loss
"""

args = ["evaluate", evaluate_command_args]
Command.run(args)


train_command_args = """
model:
    type: ssm
    checkpoint:
        type: mamba
        path: /path/to/model
trainer:
    train_iters: 1000
"""

args = ["train", train_command_args]
Command.run(args)
```

jlamypoirier: @bigximik thanks for the suggestion. It's actually very close to my [initial implementation](https://github.com/ServiceNow/Fast-LLM/pull/245/files/4a67660a2a7b10d9c953493abeddbdc501d90954), without the implicit registration. But I moved away from it following discussions here and other concerns that arose when trying it out in existing dynamic classes:
* This is opt-in from the developer perspective, but always enabled for the user. I like that but we got an explicit request to do that https://github.com/ServiceNow/Fast-LLM/pull/245#discussion_r2089805356. There is also an explicit request to have multiple registries
* Name clashes are too much of a problem. We can have a "gpt" model, but also a "gpt" trainer, a "gpt" dataset, a "gpt" huggingface model, etc. With a single global register, this is not possible and we need longer names to properly disambiguate, at which point it's basically the same as the class name. So the only viable option is to use the class name itself.
* That means long and annoying names. We can't just write "default" or "llama3" for rotary embeddings, we have to provide redundant information, ex. "Llama3RotaryConfig" or an adjusted form like "llama3_rotary".
* It's also not great for backward compatibility, since we'd have to deprecate all existing human-friendly names.
* Concerning the cli part, it goes against the [request to preserve the syntax](https://github.com/ServiceNow/Fast-LLM/pull/245#discussion_r2082570508). Your suggestion solves the "train" part (it's basically the same as mine) but leaves out the "gpt" part, so we'd still need to dispatch the actual trainer dynamically.

jlamypoirier: This is the same kind of opt-in that got rejected [here](https://github.com/ServiceNow/Fast-LLM/pull/245#discussion_r2089910687). All base classes potentially participate in the dynamic dispatch, depending on what their derived classes may do. I don't have a problem with it, but @tscholak specifically requested enabling dynamic dispatch for specific base classes only.

I'm not a fan of having a single registry with duplicate names. It works , but looks potentially confusing. I prefer an explicit specification of what base class is allowed for a given registration, it's easier to reason with and document, e.x we can make clear and explicit lists of allowed types for the base classes we care about for dynamic dispatch.

As for the cli part, your implementation would be difficult to run in practice because the actual model/trainer class is specified through inheritance rather than composition, and changing that would be really difficult. Once we take that into account, we basically get the implementation I'm proposing (see #269), independently of the actual registration mechanism.

Anyway, I've spent enough time on this an would rather move on to other tasks. #268 is fully functional and satisfies all requirements as far as I know, so I'd rather not change it anymore unless there is a very good reason to do so.

tscholak: makes sense @jlamypoirier

Issue #249, opened 2025-05-01
[bug] Documentation build hook fails


jlamypoirier:
# 🐞 Describe the Bug

It's having trouble with mamba:

```
Collecting mamba_ssm==2.2.4 (from mamba_ssm[causal-conv1d]==2.2.4; extra == "core"->fast_llm==0.2.0)
  Downloading mamba_ssm-2.2.4.tar.gz (91 kB)
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [23 lines of output]
      /opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/torch/_subclasses/functional_tensor.py:276: UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at /pytorch/torch/csrc/utils/tensor_numpy.cpp:81.)
        cpu = _conversion_method_template(device=torch.device("cpu"))
      <string>:118: UserWarning: mamba_ssm was requested, but nvcc was not found.  Are you sure your environment has nvcc available?  If you're installing within a container from https://hub.docker.com/r/pytorch/pytorch, only images whose names contain 'devel' will provide nvcc.
      
      
      torch.__version__  = 2.7.0+cu126
      
      
      Traceback (most recent call last):
        File "/opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
          main()
        File "/opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/opt/hostedtoolcache/Python/3.12.10/x64/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 175, in prepare_metadata_for_build_wheel
          return hook(metadata_directory, config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #261, opened 2025-05-12
Support of generate

### LINKED ISSUE #217: Implement  generate with the mixin from HF

bigximik:
# ✨ Description

Adds support for `generate` and extends support for `forward` without handling `cache`, `past_key_values`, `labels`, `attention` output, or `inputs_embeds`.  
`position_ids` are ignored and reconstructed from the attention mask.  
Currently, only data-parallel generation is supported.

Closes #217 

## 🔍 Type of change

Select all that apply:

- [x] 📝 **Documentation change** (updates documentation, including new content or typo fixes)

## 📝 Changes

1. Updated the `HuggingfacePreTrainedModel` interface to achieve feature parity with `FastLLMModel.from_pretrained`, and modified the `__init__` method to optionally accept `runner` and `config`.
2. Fixed a bug in attention mask creation for the non–flash attention case.
3. Modified `forward` to support `generate`, aligning behavior with Hugging Face.
4. Added tests for `generate`.
5. Added a generate usage recipe to the documentation.

## ✅ Checklist

### General

- [x] 📜 I have read and followed the [contributing guidelines](https://servicenow.github.io/Fast-LLM/developers/contributing).
- [x] 🏷️ I am using a clear and descriptive PR title that summarizes the key change or feature introduced.
- [x] 🎉 The functionality is complete, and I have tested the changes.
- [x] 📝 I have updated the documentation if needed.
- [x] ⚠️ The change does not introduce any new issues (e.g., runtime warnings, type checker errors, linting problems, unhandled edge cases).
- [x] 🧩 I have commented my code, especially in hard-to-understand areas.

### Testing

- [x] 🧪 I have added or updated tests to cover my changes.
- [x] ✔️ New and existing tests pass locally with my changes.
- [x] 🚦 I have tested these changes on GPUs and verified training stability.
- [ ] 🏋️ I have tested the changes on realistic training workloads, if applicable. ***(not applicable)***


bigximik:
Next steps are to crate tests and add docs

bigximik:
There are 3 open questions:

1. **`updated` parameter in `from_pretrained`**  
   Should the keys be tuples, or should we enable support for dot notation inside `from_pretrained` (if possible)? If not, should we wrap it in a simple dot notation parser?

2. **BatchConfig support**  
   Do we want to support the full `BatchConfig`, including `sequential_micro_batches`? Or is it sufficient as it is now, where only `micro_batch_size` is passed and the full batch size is computed as `data_parallel_groups * sequential_micro_batches`?

3. **Factory methods**  
   Should we have dedicated `from_trainer` and `from_evaluator` methods, or is the current unified `from_model` interface sufficient?

oleksost:
As a side note, to support downstream evals during training, we do not necessarily need `.generate` function, do we? 
I.e. we need it only if we want to support generation tasks (which usually take longer), but if we want to support multiple choice/likelihood based evals, we might not need it at all? (afaiu lm_eval harness is not using `.generate` for likelihood based evals?)

jlamypoirier:
> 1. **`updated` parameter in `from_pretrained`**
>    Should the keys be tuples, or should we enable support for dot notation inside `from_pretrained` (if possible)? If not, should we wrap it in a simple dot notation parser?

Let's keep the existing format and not use a new rule. `cls.config_class.from_dict` can deal with updates.

> 2. **BatchConfig support**
>    Do we want to support the full `BatchConfig`, including `sequential_micro_batches`? Or is it sufficient as it is now, where only `micro_batch_size` is passed and the full batch size is computed as `data_parallel_groups * sequential_micro_batches`?

The `micro_batch_size` is useless by itself. It's only used for pipeline parallelism and throughput computation, and both of which need the full batch config. It's ok to just raise an error for features we don't want.

> 3. **Factory methods**
>    Should we have dedicated `from_trainer` and `from_evaluator` methods, or is the current unified `from_model` interface sufficient?

I don't really see the point of `from_model`, seems like an unnecessary wrapper. `from_trainer` would be more useful since it would do actual work in gathering the necessary information from the trainer.

bigximik:
@oleksost This is a split from the larger sandbox PR for full `lm_eval` integration (#222).  
You're right — `lm_eval` does not use `generate` for some tasks, but it does for others, which we may need to support either during training or as a separate model evaluation step.

bigximik:
> Let's keep the existing format and not use a new rule. `cls.config_class.from_dict` can deal with updates.

I just added future parity with `FastLLMModel.from_pretrained`, so that I can pass updates to `from_dict`, which is called inside it.

However, I meant a different thing here — Fast-LLM's `runnable` supports dot notation to apply updates from the command line:  
https://github.com/ServiceNow/Fast-LLM/blob/denis/generate_final/fast_llm/engine/config_utils/runnable.py#L183  
The `from_dict` method on `config` does not support this, right?

That said, @tscholak is okay with using tuples as keys in `from_pretraining`, so let's leave it as implemented for now.

bigximik:
> The `micro_batch_size` is useless by itself. It's only used for pipeline parallelism and throughput computation, both of which need the full batch config. It's ok to just raise an error for features we don't want.

We still need to support different global batch sizes for the same `micro_batch_size`, which usedas to the batch size per GPU.

For example, consider a training scenario with:
- `sequence_length` = 8K  
- `global_batch_size` = 32  
- `micro_batch_size` = 1 (per GPU)  
- Model = 8B  
- GPUs = 16  

This configuration uses `sequential_micro_batches = 2` if without model parallelism because the parameters wouldn't otherwise fit on a single GPU.

However, when performing evaluation, we cannot use the same `BatchConfig`, as we do not yet support `sequential_micro_batches > 1` in inference_runner. So we can't accept the full training `BatchConfig` as-is. Instead, we need to extract just the `micro_batch_size` and multiply it by the number of GPUs to derive the global batch size for evaluation.

Additionally, `sequence_length` will differ: for `generate` or `forward`, it will not be maximum as in training, but for evaluation tasks, it varies. This means it doesn't make much sense to carry it over from training, and, if we eventually want to compute FLOPs accurately for evaluation, we’ll need to consider the actual tensor sizes in each forward pass — but that’s outside the scope of this PR.

bigximik:
> I don't really see the point of `from_model`, seems like an unnecessary wrapper. `from_trainer` would be more useful since it would do actual work in gathering the necessary information from the trainer.

Yeah, you're right — `from_model` just passes arguments one-to-one to the constructor, so I'll remove it.  
That said, I also don't see much value in adding `from_trainer` right now, since it's just a matter of passing a few additional parameters manually.

jlamypoirier:
> However, I meant a different thing here — Fast-LLM's `runnable` supports dot notation to apply updates from the command line: https://github.com/ServiceNow/Fast-LLM/blob/denis/generate_final/fast_llm/engine/config_utils/runnable.py#L183 The `from_dict` method on `config` does not support this, right?

The tuple and dot formats are basically the same, it's just parsed vs unparsed. I don't think there is much use for the unparsed version outside of a runnable, because dicts are better in-code.

> However, when performing evaluation during training, we cannot use the same `BatchConfig`, as we do not yet support `sequential_micro_batches > 1` in inference_runner. But we need to leave `micro_batch_size` = 1 to not to get OOM on GPUs. So we can't accept the full training `BatchConfig` as-is. Instead, we need to extract just the `micro_batch_size` and multiply it by the number of GPUs to derive the global batch size for evaluation.
> 
> Additionally, `sequence_length` will differ: for `generate` or `forward`, it will not be maximum as in training, but for evaluation tasks, it varies. This means it doesn't make much sense to carry it over from training, and, if we eventually want to compute FLOPs accurately for evaluation, we’ll need to consider the actual tensor sizes in each forward pass — but that’s outside the scope of this PR.

Again, `forward` and `generate` don't care about the batch config. They take a single, already preprocessed micro-batch and run it no matter the content. The `BatchConfig` we have in the inference runner is just a placeholder because the schedule runner expects one.

Note that sequential micro-batches are irrelevant for inference, it's just separate batches for all practical purposes.

bigximik:
Thanks for the details.

> Note that sequential micro-batches are irrelevant for inference — it's just separate batches for all practical purposes.

Still, if you pass a `BatchConfig` with `sequential_micro_batches > 1` to the inference runner, the `forward` will crash.

OK, I’ll remove the batch config–related parameters from the constructor.  
For now, let's just allow the `runner` to be optionally passed if we’re creating from an existing model.

bigximik:
I needed to downgrade `cairosvg` from 2.8.0 to 2.7.0 because the latest version caused issues when building the documentation, with errors occurring inside the Cairo Python wrapper.

jlamypoirier:
> I needed to downgrade `cairosvg` from 2.8.0 to 2.7.0 because the latest version caused issues when building the documentation, with errors occurring inside the Cairo Python wrapper.

Does this relate to #249?

fast_llm/engine/inference/huggingface.py
```diff
@@ -20,31 +23,74 @@ class HuggingfacePreTrainedModel(transformers.PreTrainedModel):
...
+
         # Transformers needs to be able to inspect the base model.
         self.fast_llm_base_model = fast_llm_model.base_model
-        # TODO: Support distributed models?
-        assert fast_llm_model.config.distributed.world_size == 1
 
         with transformers.modeling_utils.no_init_weights():
             self.post_init()
 
+    def forward(
```
jlamypoirier, line 68: Transformers doesn't really define an interface for it, so I don't think we need it here. (Is it even an enforced standard?)

fast_llm/engine/inference/huggingface.py
```diff
@@ -20,31 +23,74 @@ class HuggingfacePreTrainedModel(transformers.PreTrainedModel):
...
+        attention_mask: torch.Tensor | None = None,
+        position_ids: torch.Tensor | None = None,
+        past_key_values=None,
+        inputs_embeds: torch.FloatTensor | None = None,
+        labels: torch.LongTensor | None = None,
+        use_cache: bool | None = None,
+        output_attentions: bool | None = None,
+        output_hidden_states: bool | None = None,
+        return_dict: bool | None = None,
+    ) -> tuple | transformers.modeling_outputs.CausalLMOutputWithPast:
```
jlamypoirier, line 80: We can't hard-code to causal lm in this class

fast_llm/engine/inference/runner.py
```diff
@@ -42,7 +56,8 @@ def fast_llm_model(self) -> FastLLMModel:
         return self._fast_llm_model
 
     def setup(self):
-        self._runner.setup(self._fast_llm_model.distributed)
+        if not self._runner._is_setup:
+            self._runner.setup(self._fast_llm_model.distributed)
```
jlamypoirier, line 60:
```
else:
   Assert.is_(self._runner.distributed, self._fast_llm_model.distributed)
```

fast_llm/layers/language_model/head.py
```diff
@@ -180,6 +180,14 @@ def _forward_backward(
         with torch.enable_grad():
             ln_output = self.final_norm(input_)
 
+            if "output_hidden_states" in kwargs and kwargs["output_hidden_states"]:
+                # The last hidden layer output is returned normalized in the HF Transformers-style output, at least for LLama style models.
+                # So, if needed, we gather the data after normalization and set it as the output of the previous layer.
+                group = self._tensor_space.distributed.tensor_group if self._parallel_embeddings else None
+                sequence_parallel = self._sequence_parallel and self._parallel_embeddings
```
jlamypoirier, line 187: self._sequence_parallel_logits

fast_llm/layers/language_model/head.py
```diff
@@ -180,6 +180,14 @@ def _forward_backward(
...
+                hidden_state = gather_op(ln_output.detach(), group, dim=0) if sequence_parallel else ln_output.detach()
+                kwargs["hidden_states"][len(kwargs["hidden_states"]) - 1]["tensor"] = hidden_state
```
jlamypoirier, line 189: Can we make a `TensorMeta` and use `local_to_global` instead? it's a lot more testted and robust. And we already have all the information from kwargs (ex. `TransformerKwargs.hidden_dims`)

fast_llm/engine/inference/huggingface.py
```diff
@@ -54,18 +100,23 @@ def from_pretrained(
...
-        config = cls.config_class(fast_llm_model.config)
 
-        return cls(config, fast_llm_model, **kwargs)
+        return cls(fast_llm_model, **kwargs)
 
     def _init_weights(self, module) -> None:
         raise NotImplementedError(module)
+
+
+class HuggingfaceBaseModelForCausalLM(HuggingfacePreTrainedModel, transformers.generation.utils.GenerationMixin):
```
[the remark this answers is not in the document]
bigximik: Transformers only defines the `.generate()` method for **causal language model** variants — i.e., subclasses of `PreTrainedModel` that implement `PreTrainedModelForCausalLM`. For example, see the implementation in [LLaMA's `modeling_llama.py`](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L611).

Because of this, we specifically need to support `.generate()` only for causal language models.

However, I originally handled this incorrectly, as you noted in other comments:
> Transformers doesn't really define an interface for it, so I don't think we need it here. (Is it even an enforced standard?)

> We can't hard-code to causal lm in this class

So, I will keep `HuggingfaceBaseModelForCausalLM` but will move `forward`  definition which returns  `transformers.modeling_outputs.CausalLMOutputWithPast` to it from `HuggingfacePreTrainedModel`

tests/test_gpt_generate.py
```diff
@@ -0,0 +1,224 @@
...
+        hf_model,
+        fast_llm_model,
+        max_new_tokens=max_new_tokens,
+    )
+    _compare_gen_outputs(outputs, min_matching_tokens=min_matching_tokens_batch_size_2)
+
+
+@pytest.fixture(scope="module")
+def model_and_tokenizer():
+    model = "HuggingFaceTB/SmolLM2-135M-Instruct"
```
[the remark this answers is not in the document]
bigximik: Around a minute with download and around 30 sec if already downloaded

fast_llm/engine/inference/config.py
```diff
@@ -91,7 +91,8 @@ def __eq__(self, other) -> bool:
 
     def to_dict(self) -> dict[str, typing.Any]:
         out = super().to_dict()
-        out["fast_llm_config"] = self.fast_llm_config.to_dict(verbose=FieldVerboseLevel.everything)
+        if self.fast_llm_config is not None:
```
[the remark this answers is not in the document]
bigximik: done

fast_llm/engine/inference/huggingface.py
bigximik: generate expects this interface, but it must be moved to HuggingfaceBaseModelForCausalLM

fast_llm/engine/inference/huggingface.py
bigximik: generate expects this interface, but it must be moved to HuggingfaceBaseModelForCausalLM

fast_llm/engine/inference/huggingface.py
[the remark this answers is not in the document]
bigximik: done

fast_llm/engine/inference/huggingface.py
bigximik: done

fast_llm/engine/inference/huggingface.py
bigximik: done

fast_llm/engine/inference/runner.py
bigximik: done

fast_llm/layers/language_model/head.py
bigximik: done

fast_llm/layers/language_model/head.py
bigximik: done

jlamypoirier (commented):
Looks good, only need to address the documentation and slow test issues.

Concerning the doc, I pushed a tentative fix in #271, but we won't know if it works until we merge the PR.

Concerning the tests, I suggest you add one with a dummy checkpoint (ex. the one from `test_checkpoint`, and keep the existing one with a skip mark. These longer tests do make sense, and I'd like us to find a way to bring them back without disrupting existing workflows.

tscholak (approved):
LGTM!
We can postpone finding a smaller model for the test to another time. The current test should be kept but not run by default.

bigximik:
i have added extra_slow mark to tests it can be enables with `--run-extra-slow`

tests/test_gpt_generate.py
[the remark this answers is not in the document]
bigximik: added `extra_slow` mark to these tests, will be ignored by default

bigximik:
I’ve added faster tests using a dummy checkpoint from the checkpoint tests. They’re still relatively slow (around 11–15 seconds), so I’ve marked them as `slow`.

tests/test_gpt_generate_and_forward.py
```diff
@@ -0,0 +1,373 @@
...
+        use_bf16,
+        max_new_tokens,
+        min_matching_tokens_batch_size_1,
+        min_matching_tokens_batch_size_2,
+        tokenizer=tokenizer,
+    )
+
+
+@pytest.mark.slow
+@requires_cuda
```
jlamypoirier, line 261: `@pytest.mark.depends`? on the checkpoint test?

bigximik: It does not depend on it, trains a little independently and saves directly to HF format:
```python
@pytest.fixture(scope="module")
def small_model():
    from .common import _CONFIGS, TEST_RESULTS_PATH, run_test_script

    _, _, _, common_config, fast_llm_checkpoint_format = _CONFIGS["llama"]
    run_test_script(
        f"test_llama_generate_and_forward",
        common_config
        + ["training.checkpoint.interval=1", "training.export.format=llama", "training.export.interval=1"],
    )
    return TEST_RESULTS_PATH / "test_llama_generate_and_forward/export/llama/2", fast_llm_checkpoint_format
```

PR #266, opened 2025-05-14
Misc improvements and fixes

jlamypoirier:
# ✨ Description

Some misc stuff ectracted from #245 to reduce its size.
* Replace flaky and poorly formatted `__repr__` from `dataclass` with a formatted config (`to_logs`)
* Remove the ignored `default_factory` for nested config fields.
* Make instantiation consistent by forcing it to go through `_from_dict`.
* Rename `model_config_class` -> `model` in `ConvertConfig`
* Add some type hints to help static type checkers.
* Various fixes and minor improvements

## 🔍 Type of change

Select all that apply:

- [x] 🐛 **Bug fix** (non-breaking change that addresses a specific issue)
- [ ] 🚀 **New feature** (non-breaking change that adds functionality)
- [ ] ⚠️ **Breaking change** (a change that could affect existing functionality)
- [ ] 📈 **Performance improvement/optimization** (improves speed, memory usage, or efficiency)
- [x] 🛠️ **Code refactor** (non-functional changes that improve code readability, structure, etc.)
- [ ] 📦 **Dependency bump** (updates dependencies, including Dockerfile or package changes)
- [ ] 📝 **Documentation change** (updates documentation, including new content or typo fixes)
- [ ] 🔧 **Infrastructure/Build change** (affects build process, CI/CD, or dependencies)

PR #268, opened 2025-05-14
Minimalistic dynamic configs

jlamypoirier:
# ✨ Description

A simplified version of #245, with only the dynamic config support and its example usage for gpt datasets, i.e. excluding the further usage in cli and transformer sub-configs. I also made it more opt-in by making registries only on request and removing the global config registry.

## 🔍 Type of change

Select all that apply:

### DIFF
--- a/fast_llm/layers/transformer/config.py
+++ b/fast_llm/layers/transformer/config.py
@@ -95,7 +95,7 @@ class RotaryEmbeddingType(str, enum.Enum):
     yarn = "yarn"
 
 
-@config_class()
+@config_class(registry=True)
 class RotaryConfig(BaseModelConfig):
     _abstract = False
     type: RotaryEmbeddingType = Field(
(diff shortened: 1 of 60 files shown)


fast_llm/layers/transformer/config.py
```diff
@@ -244,6 +250,12 @@ def _validate(self) -> None:
                 )
 
 
+for name in PeftType:
+    # We need this because we are using the reserved field name `type`.
+    # TODO: Implement proper dynamic typing.
+    TransformerPeftConfig.register_subclass(name.value, TransformerPeftConfig)
+
```
tscholak, line 257: can we use `tag` instead of `type` to avoid this workaround?

jlamypoirier: We could but I don't think it's worth it. The workaround won't stay long since I already have dynamic implementations in #245. Also keeping `type` makes the transition easier since we're already keeping that name, eg. so config change or backward compatibility fix is needed.

tscholak (approved):
LGTM, merge when ready please

jlamypoirier:
@tscholak It's ready but depends on #266 which is waiting on review

PR #273, opened 2025-05-16
[Prototype] Run tests in parallel

jlamypoirier:
# ✨ Description

Allows running tests in parallel and using all the available gpus so we can run lots of tests fast. Pytest-xdist is already relatively good, but puts everything in the first GPU(s) and risks causing OOMs, port conflicts and other issues. I made a simple allocation and locking mechanism to prevent such issues, adapted from [pytest-xdist-lock](https://libraries.io/pypi/pytest-xdist-lock).

The system comes in a few steps:
* Test request a certain amount of gpus, gpu memory and ports through the `get_test_resources` mark or a specialized decorator such as `requires_cuda`.
* The lock adapter safely allocates the gpu(s). It sets the default device to the first allocated one and restrict gpu usage through `set_per_process_memory_fraction` (5 GB by default for requested devices, 0 for other gpus), which is good enough for many tests.
* For simple tests nothing more is needed, but more complex ones need to know the allocated gpus and ports, which they get through the `get_test_resources` fixture. This include fast-llm runs and distributed configs, for which I added config options and the `get_distributed_config` fixture, and Megatron runs which use `CUDA_VISIBLE_DEVICES`.
* Once the test is done, the lock adapter checks that the allocation was respected, ensures that the GPU memory is de-allocated, and unlock the resources for other tests.

What remains is to ensure that dependencies between tests are respected (i.e. that pytest-xdist and pytest-depends are compatible enough), and that shared resource files (ex. test dataset) are parallel-safe. 

I got things to a relatively stable state up to ~20 workers, but things start to break above it. It's still enough to reduce slow tests from 8 minutes to ~2 minutes, most of which comes from parallel overhead (~1 minute) and the slowest test (~40 s), so it adds room for lots of extra tests.

## 🔍 Type of change

Select all that apply:

- [x] 🔧 **Infrastructure/Build change** (affects build process, CI/CD, or dependencies)

### DIFF
--- a/tests/test_simple.py
+++ b/tests/test_simple.py
@@ -3,102 +3,110 @@ import pytest
 from tests.common import CONFIG_COMMON, CONFIG_FAST_LLM, TEST_MODEL, run_test_script
 
 
-def test_model_safe():
+def test_model_safe(gpu_lock):
     # The safest possible config, identical to the one in test_match_megatron except for the initialization.
-    run_test_script(
-        f"test_{TEST_MODEL}_safe",
-        CONFIG_FAST_LLM
-        + [
-            "run.torch_dynamo_enable=False",
-            "schedule.data_overlap=False",
-            "model.base_model.transformer.dropless_moe=False",
-        ],
-    )
+    with gpu_lock():
+        run_test_script(
+            f"test_{TEST_MODEL}_safe",
+            CONFIG_FAST_LLM
+            + [
+                "run.torch_dynamo_enable=False",
+                "schedule.data_overlap=False",
+                "model.base_model.transformer.dropless_moe=False",
+            ],
+        )
 
 
 @pytest.mark.depends(on=["test_model_safe"])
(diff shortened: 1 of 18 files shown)


tests/test_simple.py
```diff
@@ -3,102 +3,110 @@
...
-def test_model_tp2():
+def test_model_tp2(gpu_lock):
     # Simple tensor-parallel.
-    run_test_script(
-        f"test_{TEST_MODEL}_tp2",
-        CONFIG_COMMON + ["model.distributed.tensor_parallel=2"],
-        num_gpus=2,
-        compare=f"test_{TEST_MODEL}",
-    )
+    with gpu_lock(num_gpus=2):
```
jlamypoirier, line 69: Ideally I'd use a mark or a better fixture to avoid this. Any pytest expert here knows how to do it? (@bigximik @tscholak)

PR #275, opened 2025-05-27
Dynamic transformer

jlamypoirier:
# ✨ Description

Adjust the rotary embeddings, peft and normalization layers to use the new dynamic classes. Do some cleanup and refactoring for rotary embeddings. Add an option to disable normalization layers because why not.

## 🔍 Type of change

Select all that apply:


fast_llm/layers/transformer/rotary/rotary.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def _get_frequencies(self, sequence_length: int, kv_channels: int, device="cuda") -> torch.Tensor:
+        return super()._get_frequencies(sequence_length, kv_channels, device) * self._config.attention_factor
+
+    def _get_angle_scales(self, kv_channels: int, device="cuda") -> torch.Tensor:
+        scales = super()._get_angle_scales(kv_channels, device)
+        # TODO: max_position_embeddings or original_context_length?
+        # see https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/modeling_deepseek.py#L304
+        low = max(self._get_correction(self._config.beta_slow, kv_channels), 0)
+        high = min(self._get_correction(self._config.beta_fast, kv_channels), kv_channels - 1)
```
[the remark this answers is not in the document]
RaymondLi0: Hmm, well the original implementation uses 
- `beta_fast` to compute `low` 
- `beta_slow` to compute `high`

fast_llm/layers/transformer/rotary/rotary.py
```diff
@@ -0,0 +1,214 @@
...
+    def _linear_ramp_factor(self, min, max, dim):
+        if min == max:
+            max += 0.001  # Prevent singularity
+
+        linear_func = (torch.arange(dim, dtype=torch.float32) - min) / (max - min)
+        ramp_func = torch.clamp(linear_func, 0, 1)
+        return ramp_func
+
+    def _get_correction(self, beta: float, dim: int) -> float:
+        return math.floor(
```
RaymondLi0, line 210: original implementation uses `floor` for `low`, but `ceil` for `high`.

fast_llm/layers/transformer/rotary/rotary.py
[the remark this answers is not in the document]
jlamypoirier: You're right, sorry for the confusion

fast_llm/layers/transformer/config.py
```diff
@@ -194,66 +160,50 @@ class TransformerPeftConfig(PeftConfig):
     )
```
oleksost, line 160: Not sure this is the right place to raise this concern, but I think its better to not have layer freezing as part of PEFT, and keep only a single layer freezing mechanism (e.g. currently using lr sclaing for different components and layers).  

Hence, if user wants to use PEFT (e.g. LoRA) in conjunction with layer freezing, they might do it by explicitly freezing the layers using `lr_scale` parameter. 

This way we can keep the LoRA/PEFT logic simple and disentangle parameter freezing from PEFT.

fast_llm/layers/common/config.py
```diff
@@ -120,27 +136,47 @@ def _from_dict(
...
+
+    @property
+    def module_class(self):
+        from fast_llm.layers.common.normalization import RMSNorm
+
+        return RMSNorm
 
 
 @config_class()
 class PeftConfig(BaseModelConfig):
```
oleksost, line 162: no default handling (from_dict)?

jlamypoirier: We don't really use that one and it doesn't have a registry, these are in `TransformerPeftConfig` instead.

fast_llm/layers/transformer/config.py
jlamypoirier: That's a fair point, but LoRA is almost always used together with freezing, so it's really more convenient to do it together. Also I don't think the arbitrary lr scaling parameters is a good long term solution, and anyway there is still only one freezing mechanism in the background, so it's not too bad.

Issue #242, opened 2025-04-24
Support block-modular architecture


tscholak:
# 🎯 **Goal (What & Why)**

Enable fully **modular, per-block configuration** in Fast-LLM to follow up on hybrid architecture support introduced in #194.

Currently, hybrid models (e.g., interleaving Mamba 1, Mamba 2, and Transformer blocks) are limited by **global block-type configurations**: all transformer blocks share one config, and all SSM blocks another. This is too rigid.

We want to:
- Allow **different configurations per block**, even for the same type.
- Support **named blocks** with configurable **weight sharing**.
- Enable expressive, fine-grained **architectures**, useful for:
  - Exploring different attention mechanisms in a single model.
  - Tying weights across repeated block instances.
  - Designing sparse, pruned, or ablation-based stacks.
  - Preparing for model export pipelines with heterogeneous block stacks.

This would eliminate the current one-size-fits-all limitation and make model stacks in Fast-LLM truly composable and expressive.

# 🚀 **Execution Plan**

This is a config and model-construction feature. The main change is replacing the global `transformer` and `ssm` sections with a new per-block format.

### Key Ideas

- Add `model.blocks`: a dict of named block configs (e.g., `alice`, `bob`, `claire`, `potato`, etc., it doesn't matter what they are called, see example below).
- Add `block_pattern`: a list specifying the block sequence by name.
- Add `num_layers`: total depth of the model. The pattern repeats to reach this.
- Allow block-level options like:
  - `kind: transformer | ssm | ...`
  - `attention_type`, `sliding_window`, `dt_rank`, etc.
  - `shared_weights: true` for parameter sharing
  - `lora: ...`
- Blocks reused by name will share configuration; if `shared_weights: true`, they’ll also reuse parameters.

### Minimal Implementation Path

1. Define new schema and validate it (e.g., every pattern entry must resolve to a block).
2. Update model construction to instantiate blocks from `model.blocks`, repeat pattern to reach `num_layers`.
3. Add weight-sharing logic: instantiate shared blocks once, reuse parameters across layers.
4. Add support for block-level LoRA injection.
5. Maintain backwards compatibility: for existing models, fall back to current global `transformer`/`ssm` layout if `model.blocks` is absent. Save new checkpoints using the new format.
6. Extend test coverage:    - Stacks with different transformer configs
    - Mixed MQA/GQA/sliding-window blocks
    - Interleaved SSM and transformer blocks
    - Shared and unshared weights
[...]

PR #277, opened 2025-05-29
[Prototype] Support block-modular architecture

oleksost:
# ✨ Description

This draft PR addresses #242 by introducing a flexible, modular configuration system for hybrid model architectures.

TODOs:
- [ ] merge `main` into this branch
- [ ] add more testing to make sure legacy behaviour is well supported
- [x] implement weight sharing
- [x] support for block-specific learning rate scales
- [ ] make sure model serialisation/conversion works as expected
- [ ] review and unify naming conventions (block, layer) across codebase.
- [ ]  clean up & test

```
model:
  base_model:
    cross_entropy_impl: fused
    blocks:
      bob:
        type: transformer
        hidden_size: 512
        share_weights: true
        
      mamba:
        type: discrete_mamba2
        state_size: 16
        expansion_factor: 2
        hidden_size: 512
        
    hybrid_block_layout: ["bob", "mamba", "mamba", "bob"]
    num_layers: 4
```

Which will result in block layout like this: ["bob", "mamba_1", "mamba_2", "bob"], where bobs share weights and mamba do not share weights.

## 🔍 Type of change
Class hierarchy in the config system:
- started moving functionality specific to `BaseBlock` into `BaseBlockConfig` in `layers/common`
- transformer and SSM layer configs inherit from `BaseBlockConfig`, both holding functionality specific to their dedicated blocks (TransformerLayer, LlambaBlock)

**Block-specific hyperparameters & tensor space definition:**
- `HybridBlockConfig`s implemented under `models/hybrid/config` allowing block-specific hyperparameters definition
- the names of the elements in the tensor space now include block suffixes; no suffixes are used in the case of non-hybrid GPT models
- still supports legacy behaviour with blocks defined using lists like [t,m2d,m] & non-hybrid GPT models

**Layer freezing**:
- in case of PEFT layer freezing **must** be explicit: i.e. if LoRA is used, it dpoes nto automatically freeze other layers and lr_scales must be used.
- we have per_block `lr_scale` and component specific scales like `norm_lr_scale`, `mlp_lr_scale` etc.. If both are passed, the resulting scale for a component is `lr_scale` of the block multiplied by the component specific lr (see 'get_lr_scale' function.
- for GPT model (non-hybrid) 'lr_scale' should not be used as it would be applied to all layers, since all alyers share the same config.

Select all that apply:

- [ ] 🐛 **Bug fix** (non-breaking change that addresses a specific issue)
- [x] 🚀 **New feature** (non-breaking change that adds functionality)
[...]

PR #339, opened 2025-07-29
Block interface: refactor

jlamypoirier:
# ✨ Description

Decouple the block interface from the transformer. This will make the SSM interface cleaner, improve readability, prevent bug, simplify the implementation of future mixers, etc. It's also a step towards merging SSMs into the GPT model, and varying block configurations (#242). I also included some groundwork for these tasks, so that upcoming PRs are smaller and simpler.

This PR is pure refactor, and changes that would cause backward compatibility concerns are left for future work.

* Create a `block` submodule and move everything non-specific to transformers there from the `transformer` directory.
* Extract base classes `BlockConfig`, `MLPConfig`, `AttentionConfig` from `TransformerConfig`. Using inheritance for backward compatibility, but composition would be preferable.
* Extract base classes `BlockDimNames`, `BlockKwargs` from their transformer counterparts for common variables. Use inheritance in their specializations to simplify usage. (Ex. we can systematically use `SSMDimNames`, `SSMKwargs` when dealing with SSMs.)
* Remove `TensorSpace` entirely, instead create tensor dimensions on the fly. This will both make things simpler and prevent unwanted conflicts between layers with different configs.
* Remove unused fields `per_layer_lr_scale`(from incorrect inheritance of `BlockConfig`) and `normalization` from `SSMConfig`, which would cause unexpected behaviour if defined.
* Make a layer debugging utility to merge the various debug calls and help add new ones.
* Move initialization to its own file.
* Lots of misc interface fixes and improvements.

## 🔍 Type of change

Select all that apply:


tscholak:
@jlamypoirier, thanks for pushing on this. Unifying GPT and HybridSSM is absolutely the right long-term goal, and this PR on block abstractions is valuable groundwork.

Before we get too deep into this, I want to clarify expectations on the direction for the block architecture.

Eventually, we need to land the feature #242 exactly as scoped: declarative, per-block config, explicit architecture, full heterogeneity, and named block/weight sharing. Please have a look also at #277.

That's the path forward we need. Anything that diverges from this or locks us into an alternative config approach (including any override machinery revival) is off the table and won't be accepted by me or the team.

Your skills are exactly what we need to lay the groundwork for #242 quickly and cleanly. That doesn't mean you need to implement the whole feature yourself. What matters is that your work enables us to land #242 without obstacles.

Thanks.

jlamypoirier:
@tscholak I'm aiming quite close to the scope https://github.com/ServiceNow/Fast-LLM/pull/342/files#diff-a600ef5264541f7a4a8b245115233343926edacce94e81249e31134e8760adf1R174.

After having a look, I think it's safer that I implement it myself. There are very complex backward compatibility implications to deal with, plus it's a good time to deal with some old technical debt and add much-needed improvements. 

I'll leave out weight sharing for now though because it's a brand new feature and is not directly related.

tscholak (approved):
LGTM!

PR #340, opened 2025-07-30
[WIP] Add tensor parallelism (and general model/sequence parallelism) support for HF wrapper forward and lm_eval integration

### LINKED ISSUE #334: Support Tensor Parallelism in inference
# 🎯 **Goal (What & Why)**
We need support for TP (tensor parallelism) in inference to enable evaluators to run during training with TP enabled. This is needed sooner than full support for model and sequence parallelism (#278), which is why it's tracked as a separate issue.

bigximik:
# ✨ Description

Add tensor parallelism (and general model/sequence parallelism) support for HF wrapper forward and lm_eval integration

Closes #334 

## 🔍 Type of change

Select all that apply:

## 📝 Changes

1. Change A
2. Change B

## ✅ Checklist

### General

- [ ] 📜 I have read and followed the [contributing guidelines](https://servicenow.github.io/Fast-LLM/contributing/contributing).

### Testing

### Performance Impact

---

## 🗒️ Additional Notes

### DIFF
--- a/fast_llm/engine/inference/huggingface.py
+++ b/fast_llm/engine/inference/huggingface.py
@@ -13,7 +13,6 @@ from fast_llm.engine.inference.runner import InferenceRunner
 from fast_llm.engine.multi_stage.config import StageMode
 from fast_llm.engine.multi_stage.fast_llm_model import FastLLMModel
 from fast_llm.engine.schedule.runner import ScheduleRunner
-from fast_llm.utils import Assert
 
 logger = logging.getLogger(__name__)
 
@@ -56,10 +55,6 @@ class HuggingfacePreTrainedModel(transformers.PreTrainedModel):
         # A model can be created from pretrained which set it up in the current HF wrapper api
         # or set existing model which  also must be setup, so, do not accept not setup model
         assert fast_llm_model.is_setup
-
-        # We only support data parallel for now
-        Assert.eq(fast_llm_model.distributed.config.model_parallel, 1)
-        Assert.eq(fast_llm_model.distributed.config.sequence_data_parallel, 1)
 
         self._inference_runner.setup()
 
--- a/fast_llm/layers/language_model/head.py
+++ b/fast_llm/layers/language_model/head.py
@@ -267,6 +267,9 @@ class LanguageModelHead[ConfigType: LanguageModelBaseConfig](Configurable[Langua
             )
             if targets is None:
                 # TODO: Make a proper way of returning the model output.
+                if "global_logits" in kwargs and kwargs["global_logits"] == True:
+                    logits_meta = self._get_logits_meta(kwargs, loss)
+                    loss, _ = logits_meta.local_to_global(loss.detach(), distributed=self._tensor_space.distributed)
                 kwargs["logits" if self._prediction_distance == 0 else f"logits_{self._prediction_distance}"] = loss
                 return None, None
         else:
@@ -306,6 +309,19 @@ class LanguageModelHead[ConfigType: LanguageModelBaseConfig](Configurable[Langua
             all_reduce(loss, group=self._tensor_space.distributed.tensor_group)
         return loss, logit_input_grad.view_as(input_) if logit_input_grad is not None else None
 
+    def _get_logits_meta(self, kwargs: dict, logits: torch.Tensor, logits_name: str = "transformer logits"):
+        vocab_dim = self._tensor_space.get_tensor_dim(
+            LanguageModelDimNames.vocab if self._sequence_parallel_logits else LanguageModelDimNames.vocab_tp
+        )
+        dims = [*kwargs[TransformerKwargs.hidden_dims][:-1], vocab_dim]
+        sequence_index = 1 - int(kwargs[TransformerKwargs.sequence_first])
+        dims[sequence_index] = (
+            TensorDim(TransformerDimNames.sequence_q_tp, dims[sequence_index].global_size, DistributedDimNames.tensor)
+            if self._sequence_parallel_logits
+            else TensorDim(TransformerDimNames.sequence_q, dims[sequence_index].global_size)
+        )
+        return TensorMeta.from_dims(tuple(dims), tensor_name=logits_name, dtype=logits.dtype)
+
     def _logits_cross_entropy_forward_backward(
         self,
         input_: torch.Tensor,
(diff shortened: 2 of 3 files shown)


fast_llm/engine/evaluation/config.py
```diff
@@ -98,6 +98,13 @@ class LmEvalEvaluatorConfig(EvaluatorConfig):
         " If not set, it is inferred from the Fast-LLM model config or tokenizer.",
     )
 
+    communication_timeout_sec: float = Field(
```
jlamypoirier, line 101: `timeout`. Unnecessary long timeouts are often bad, so I recommend making it optional (default none) and enabling only as needed.

fast_llm/engine/inference/huggingface.py
```diff
@@ -127,3 +129,148 @@ def forward(
     ) -> tuple | transformers.modeling_outputs.CausalLMOutputWithPast:
         # Meant to be overridden in derived classes
         raise NotImplementedError()
+
+    def forward(
```
jlamypoirier, line 133: This doesn't seem relevant outside lm_eval. Any way to move it there?

fast_llm/layers/language_model/head.py
```diff
@@ -273,6 +273,14 @@ def _logits_cross_entropy_forward_backward_split(
             )
             if targets is None:
                 # TODO: Make a proper way of returning the model output.
+                loss = loss.detach()
```
jlamypoirier, line 276: Simplified this in merge with main

fast_llm/engine/inference/huggingface.py
bigximik: I initially thought about handling this differently, but since each subclass of the model has its own class, the only practical way I found was to use a **dynamic class** that constructs itself on the fly with `type`.  
This lets us encapsulate `forward` of the `fast_llm` Hugging Face interface and then pass it to `generate`.  

Something like:

```python
def wrap_hf_model(model):
    inner_forward = get_bounded_method(model.forward, model)
    wrapper_class = get_new_type(
        model.__class__,
        {
            "inner_forward": inner_forward,
            "forward": cordinator_forward,
            "worker_forward": worker_forward,
        },
    )
    model.__class__ = wrapper_class
    return model
 ```

Another option would be to create a **static wrapper class**, but that would require exposing and forwarding a lot of functionality that ""generate"" expects.

So instead, I decided to implement this in our **HF wrapper**, since it executes before any class specialization.

File: fast_llm/engine/evaluation/config.py (hunk 1 above)
Comment (bigximik) [reply fast_llm/engine/evaluation/config.py:101]: ### Context

Conceptually, places like `worker_forward` or `data-parallel_worker` wait primitives should only exit under three conditions:

1. They receive work  
2. They receive a finish message  
3. The connection with peers/coordinator is lost (after some timeout)  

However, this is not how `torch.distributed` works. It is designed for  more or less **synchronous** communication, while here we are trying to adapt it for **asynchronous** communication.

### Problem

If we set the default timeout to `None`, users will end up seeing random timeouts in different places.

### Discussion

A better long-term solution would be to use a distributed messaging framework that is more appropriate for sending work and finish messages. However, introducing another communication layer into `fast_llm` is likely outside the scope of this PR.

### Proposal

- Keep the default timeout as it is, applied only to these entry points. reset timeout after wait operation to default of 60 sec.
- Clarify the naming/description to avoid confusion.  
- Add a TODO to revisit this later with a more suitable communication framework.

File: fast_llm/engine/inference/huggingface.py (hunk 2 above)
Comment (bigximik) [reply fast_llm/engine/inference/huggingface.py:133]: No, `lm_eval` calls `FastLLMLmEvalWrapper`, which then calls our `model.generate` which in turn makes multiple calls to `forward`, all runs entirely on TP rank 0. Our `forward` must be overridden to handle data distribution across all TP ranks.

```
----------------------------------------------------------------------------------------------------
|                                         Tensor Parallel Setup                                    |
----------------------------------------------------------------------------------------------------

   [ lm_eval ]                   [ forward_worker ]                [ forward_worker ]   ... more
        |                                 |                                  |
        v                                 v                                  v
+-------------------------+               |                                  |
| HF generate mixin       |               |                                  |
| model must  be HF       |               |                                  |
| (model.generate)        |               |                                  |
| - runs only on TP rank0 |               |                                  |
| - does multiple forward |               |                                  |
|   calls + sampling      |               |                                  |
|   beam search, etc.     |               |                                  |
+-------------------------+               |                                  |
        |                                 |                                  |
        v                                 |                                  |
+-------------------------+               |                                  |
| model.forward()         |---------------+----------------------------------+--> [wait for data]
| must be overridden      |               |                                  |     [long timeout]
| - orchestrates TP calls |               |                                  |     
+-------------------------+               |                                  |
        |                                 |                                  |
        v                                 v                                  v
+---------------------+       +---------------------+          +---------------------+   ... TP N-1
| TP rank 0           |       | TP rank 1           |          | TP rank 2           |
| model.forward_inner |       | model.forward_inner |          | model.forward_inner |
+---------------------+       +---------------------+          +---------------------+
```

Alternatives i have considered:

1. **Composition (wrapper-model around HF specialization)**

- Define a class `OrchestratorModel` that looks like an HF model (has `forward`, `generate`, etc.).
- It contains an inner `HuggingfaceBaseModelForCausalLM` (or subclass) that runs on workers.
- Orchestrator `forward` does the TP dispatch/gathering, then calls into the inner worker models as needed.
- `generate` (from HF) runs on this outer orchestrator class, which works because it just calls `self.forward`.

This is clean, explicit, stable — but involves boilerplate to replicate the HF interface.

2. **Dynamic class injection (multiple inheritance / runtime patching)**

- Build a class at runtime that inherits both:
  - HF specialization (`HuggingfaceGPTModelForCausalLM`, etc.)
  - `OrchestratorMixin` (overrides `forward`).
- Register that as the actual model class for the model object.
- `generate` is inherited unmodified from HF mixin, but calls `OrchestratorMixin.forward`.

This avoids extra wrapper code, but is “hackier” and could break with HF updates.

That’s why I dismissed the other options, but if we really want to keep `HuggingfaceBaseModelForCausalLM `completely unmodified, option 1 (composition) is likely the safer and more maintainable approach for our use case.

jlamypoirier (approved):

PR #366, opened 2025-09-22
[Prototype] Concatenated weights and linear layers

jlamypoirier:
# ✨ Description

Handle concatenated weights, i.e. weights that are stored as one in memory for computation / optimization purpose but actually consist of multiple logically distinct ones. Ex. Key and value, gated MLP layer 1, MoE layers, SSM in_proj, etc.

This lets the engine know about the structure, and allow configuring each sub-parameter independently (lr_scale, initialization, weight decay, peft enabling. Bias enabling split possible but not implemented).

Prototype is working but postponing finalization in favor of higher priority content. Remaining:
* Simplify linear layer concatenation and make it safer (especially peft)
* Use for remaining linear layers
* Make remaining tests pass.

### DIFF
--- a/fast_llm/engine/config_utils/parameter.py
+++ b/fast_llm/engine/config_utils/parameter.py
@@ -10,29 +10,10 @@ if typing.TYPE_CHECKING:
     from fast_llm.tensor import ParameterMeta
 
 
-def combine_lr_scales(*lr_scales: float | None | tuple[float | None, ...]):
+def combine_lr_scales(*lr_scales: float | None) -> float | None:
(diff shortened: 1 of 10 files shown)


jlamypoirier:
Claude Opus 4.8 note: Closing this prototype in favor of #533, which captures the goal and the prototype's mechanism (the `ConcatenatedParameterMeta` approach) at design level. The branch is ~108 commits behind `main` and predates the linear/PEFT rework, so it's preserved as a reference rather than a rebase target. Implementation details would be stale by the time this is picked up.

PR #369, opened 2025-09-26
Vision multimodal

jlamypoirier:
# ✨ Description

An attempt at integrating multimodal vision models to main. Still a lot of work to do...

PR #370, opened 2025-09-26
Base model interface review

jlamypoirier:
# ✨ Description

A series of changes extracted from #369 centered around the base model interface, aimed to make it simpler, more intuitive (closer to plain pytorch) and more flexible.

Note that this introduces additional breaking changes on checkpoints (model structure / parameter names) and configs (minor tweaks), assuming it's ok given that we just broke everything anyway, but could mitigate if needed.

Notable changes (details coming):
* Relax requirement that the base model be a flat sequence of layers. The model can now have any module structure, instead it is required to provide a list of layers to the Fast-LLM engine. This will help sort out complex models and make things cleaner. 
* Promote the decoder to an actual module (enabled by above). This improves the config/module parity and allows moving functionality (ex. preprocesing) to the module side. Add kwarg namespaces to fix name conflict issues in preprocessing.
* Promote MTP to an independent dynamic head type. This was made necessary by the recent changes, and the existing workarounds were becoming unsustainable. New format has its own block config and one block per head (now including the first one). It's a bit more complicated than before but basically unavoidable.
* Change language model / gpt structure accordingly. Replace flat list of submodules with appropriately named `embedding`, `decoder`, `head`. Use matching names for the associated config parameters
* Drop preprocessors entirely. Move model preprocessing to the associated modules. Move preference spans directly to gpt preprocessing (it's batch splitting, will need separate work). This way preprocessing is as close as possible to the code it's preprocessing for.
* Move `get_loss_definitions` to module instances (same reason).
* Simplify tied weight interface. Tied weights are now defined and used as any other, and the model provides the set of tied weights, Fast-LLM engine handles the rest.
* Make mlps and mixers respect the `Block` interface. No immediate benefit, but will allow using the mlp directly for the vision adapter, and opens up new possibilities.

Also brought minor things from #369 to reduce diff.


fast_llm/layers/attention/attention.py
```diff
@@ -429,3 +436,115 @@ def get_compute_usage(self, input_: TensorMeta, kwargs: dict[str, typing.Any], c
                 self.dense.get_compute_usage(dense_input, config),
             )
         )
+
+    def preprocess(self, batch: torch.Tensor, kwargs: dict[str, typing.Any]) -> None:
+        self._rotary.preprocess(batch, kwargs)
+        if not self._use_flash_attention:
```
tscholak, line 442: could we make this a config option, an enum? we select other implementations like that

jlamypoirier: It's already an option. We could turn into an enum, but I don't see too much reason

jlamypoirier: It would be useful if we could select the varlen version this way. That would need extra work because it's currently determined in `BatchConfig.cross_document_attention` and affects data processing, so won't do now but that's something to consider for the future.

fast_llm/layers/language_model/head.py
```diff
@@ -436,14 +429,83 @@ def _logits_cross_entropy_forward_backward(
...
+        return name
+
+    @functools.cached_property
+    def _distillation_loss_name(self) -> str:
+        name = "distillation_loss"
+        if self._prediction_distance > 0:
+            name = f"{name}_{self._prediction_distance}"
+        return name
+
+    def get_loss_definitions(self, count: int = 1) -> list[LossDef]:
```
tscholak, line 475: this is just for bookkeeping and logging?

fast_llm/layers/language_model/head.py
```diff
@@ -436,14 +429,83 @@ def _logits_cross_entropy_forward_backward(
...
         if self.training and losses is not None:
             if dpo_loss is not None:
-                losses[LanguageModelLossNames.dpo_loss].append(dpo_loss.detach())
+                losses[self._dpo_loss_name].append(dpo_loss.detach())
             if self._config.distillation_model is not None and distillation_loss is not None:
-                losses[LanguageModelLossNames.distillation_loss].append(distillation_loss.detach())
+                losses[self._distillation_loss_name].append(distillation_loss.detach())
             if self._config.distillation_model is not None and lm_loss is not None:
-                losses[LanguageModelLossNames.distil_lm_loss].append(lm_loss.detach())
+                losses[self._distillation_language_model_loss_name].append(lm_loss.detach())
```
tscholak, line 436: it would help to clarify the purposes of the 1) detached losses, 2) loss definitions, and 3) non-detached returned loss.
1) and 2) are 1-on-1 related, but idk if 3) and 2) are.

fast_llm/models/gpt/model.py
```diff
@@ -35,24 +34,43 @@ def __init__(
...
                 )  # Noqa
-        # `self._reference_models` is not populated at this point, so we pass a mutable dict.
-        self._preprocessors: list[Preprocessor] = self._config.get_preprocessors(distributed_config)
 
-    def get_layers(self) -> list[Layer]:
-        return self._config.get_blocks(self._distributed_config)
+    def get_layers(self) -> list["Layer"]:
+        return self.embeddings.get_layers() + self.decoder.get_layers() + self.head.get_layers()
 
     def preprocess_meta(
```
tscholak, line 70: remove?

jlamypoirier: Not possible atm, it's no longer in the preprocessor/layer interface but is still there for base models. Planning to revisit this with the batch-splitting mechanism, but leaving for later.

tscholak (approved):

PR #372, opened 2025-10-06
Language model block

jlamypoirier:
# ✨ Description

A small follow-up to #370, turning the language model itself into a module just like any other. Also move the hidden size to the language model config to enforce that module input/output dimensions are set by their parent module.

Also added small safety tweaks.

### DIFF
--- a/tests/utils/model_configs.py
+++ b/tests/utils/model_configs.py
@@ -344,6 +344,8 @@ _update_and_add_testing_config(
         ModelTestingGroup.distributed: ModelTestingGroupAction.unimportant,
     },
 )
+del MODEL_CONFIGS["starcoder_2"].config_dict["model"]["base_model"]["embeddings"]["num_position_embeddings"]
+
 
 _update_and_add_testing_config(
     # Main tested model.
--- a/fast_llm/engine/config_utils/parameter.py
+++ b/fast_llm/engine/config_utils/parameter.py
@@ -1,7 +1,8 @@
 import math
 import typing
 
-from fast_llm.config import Config, Field, FieldHint, config_class
+from fast_llm.config import Field, FieldHint, config_class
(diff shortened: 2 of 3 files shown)


fast_llm/engine/checkpoint/huggingface.py
```diff
@@ -170,7 +169,7 @@ def _load_weights(
                     for key in f.keys():
                         yield key, "weights", f.get_slice(key)
             elif path.suffix == ".bin":
-                # TODO: Prevent unsafe by default
-                yield from torch.load(path)
+                # TODO: Confirm that loading works with `weights_only=True`
+                yield from torch.load(path, weights_only=True)
```
tscholak, line 173: thanks for making our sec folks happy

tests/utils/model_configs.py
```diff
@@ -344,6 +344,8 @@ def _update_and_add_testing_config(
         ModelTestingGroup.distributed: ModelTestingGroupAction.unimportant,
     },
 )
+del MODEL_CONFIGS["starcoder_2"].config_dict["model"]["base_model"]["embeddings"]["num_position_embeddings"]
```
tscholak, line 347: ?

Dockerfile
```diff
@@ -47,3 +47,8 @@ COPY --chmod=777 ./tests tests
 COPY --chmod=777 ./tools tools
 COPY --chmod=777 ./fast_llm_external_models fast_llm_external_models
 COPY --chmod=777 --exclude=./fast_llm/csrc/ ./fast_llm/ fast_llm/
+
+# Set a dummy default user so we don't run in root by default.
+# The image is still compatible with any user id.
+RUN useradd user
+USER user
```
tscholak, line 54: thanks!

tscholak (approved):
LGTM!

PR #373, opened 2025-10-12
Add stochastic mixer for supernet training

tscholak:
## Summary

Implements a stochastic mixer layer for supernet training, enabling random sampling from multiple mixer options (e.g., attention vs. Mamba) during training. Includes checkpoint conversion support and a hierarchical beam search tool for finding optimal mixer placement post-training.

### Stochastic Mixer (`fast_llm/layers/decoder/stochastic_mixer.py`)
- **Training mode**: Randomly samples from configured mixers using distributed RNG (ensures consistency across TP/PP ranks)
- **Eval/inference mode**: Uses `main_mixer_index` for deterministic behavior
- **Sampling strategies**: Uniform or weighted sampling
- **Preprocessing**: Runs preprocessing for all mixers since we don't know which will be selected

### Configuration (`fast_llm/layers/decoder/config.py`)
- `StochasticMixerConfig`: List-based mixer configuration with sampling strategy
- `main_mixer_index`: Specifies which mixer to use during inference and which receives pretrained weights during checkpoint conversion
- Validation ensures sampling weights sum to 1.0 and all indices are valid

### Checkpoint Conversion (`fast_llm/models/gpt/conversion/apriel.py`)
- `AprielStochasticMixerConverter`: Handles conversion between Fast-LLM and Apriel formats
- Only `main_mixer_index` weights are exported/imported (other mixers randomly initialized during supernet training)
- Follows existing converter patterns (minimal, no verbose comments)

### Beam Search Tool (`tools/supernet_beam_search.py`)
- **Hierarchical algorithm**: Finds optimal placement for mixers at each quality/cost level
  - Phase 1: Find best N layers for primary mixer (e.g., full attention)
  - Phase 2: Find best M layers for secondary mixer (e.g., sliding window attention)
  - Remaining layers use tertiary mixer (e.g., linear attention)
- **Efficient evaluation**: Loads checkpoint once, modifies `main_mixer_index` in-place for each candidate
- **Fast-LLM integration**: Uses Fast-LLM's evaluation system directly (no subprocess or checkpoint reconversion)
- **Features**: Pre-scoring, beam growth, early stopping, configurable score direction

### Tests (`tests/utils/model_configs.py`)
- Added `stochastic_mixer` test configuration with FA/Mamba mixers
- Enabled checkpoint conversion testing via `AprielHybridSSMCheckpointFormat`

## Use Case

**Supernet Training**: Train a model where each layer can be either full attention or Mamba, with random sampling at each step. After training, use beam search to find which specific layers benefit most from full attention vs. Mamba, given a budget constraint (e.g., "I can afford 4 FA layers").

## Testing

Run the stochastic mixer tests:
```bash
pytest tests/models/test_checkpoint.py::test_checkpoint_and_eval tests/models/test_checkpoint.py::test_conversion -k "stochastic_mixer" -v
```

Example beam search usage:
```bash
fast-llm tools/supernet_beam_search.py \
  training_config=path/to/supernet_config.yaml \
  budgets=[4,8] \
  beam_width=12 \
  score_metric="lm_eval/accuracy" \
  output_path=results.json
```

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/fast_llm/layers/decoder/config.py
+++ b/fast_llm/layers/decoder/config.py
@@ -4,7 +4,7 @@ import typing
 from fast_llm.config import Field, FieldHint, check_field, config_class
 from fast_llm.engine.config_utils.parameter import combine_lr_scales
 from fast_llm.engine.config_utils.tensor_dim import TensorDim
-from fast_llm.engine.distributed.config import DistributedConfig
+from fast_llm.engine.distributed.config import _BIG_PRIMES, DistributedConfig
 from fast_llm.layers.block.config import BlockConfig, BlockKwargs
 from fast_llm.layers.common.normalization.config import NormalizationConfig
 from fast_llm.layers.common.peft.config import PeftConfig
@@ -146,6 +146,12 @@ class StochasticMixerConfig(MixerConfig):
         "and checkpoint saving (only this mixer is exported). "
         "If None, uses the first mixer in the dict.",
         hint=FieldHint.feature,
+    )
+
+    seed_shift: int = Field(
+        default=_BIG_PRIMES[11],
+        desc="Seed shift for mixer sampling reproducibility.",
+        hint=FieldHint.optional,
     )
 
     def _validate(self) -> None:
--- a/fast_llm/layers/decoder/stochastic_mixer.py
+++ b/fast_llm/layers/decoder/stochastic_mixer.py
@@ -9,7 +9,11 @@ from fast_llm.engine.distributed.config import DistributedConfig
 from fast_llm.engine.distributed.distributed import Distributed
 from fast_llm.layers.common.peft.config import PeftConfig
 from fast_llm.layers.decoder.block import BlockWithBias
-from fast_llm.layers.decoder.config import StochasticMixerConfig, StochasticMixerKwargs, StochasticMixerSamplingStrategy
+from fast_llm.layers.decoder.config import (
+    StochasticMixerConfig,
+    StochasticMixerKwargs,
+    StochasticMixerSamplingStrategy,
+)
 from fast_llm.tensor import TensorMeta
 
 logger = logging.getLogger(__name__)
@@ -118,11 +122,13 @@ class StochasticMixer[ConfigType: StochasticMixerConfig](BlockWithBias[ConfigTyp
         return self.mixers[mixer_name]._forward(input_, kwargs, losses, metrics)
 
     def preprocess(self, kwargs: dict[str, typing.Any]) -> None:
+        from fast_llm.engine.distributed.config import MAX_SEED
         from fast_llm.layers.block.config import BlockKwargs
 
         iteration = kwargs[BlockKwargs.iteration]
         generator = torch.Generator(device="cpu")
-        generator.manual_seed(iteration)
+        seed = (self._distributed_config.seed + self._config.seed_shift + iteration) % MAX_SEED
+        generator.manual_seed(seed)
         kwargs[StochasticMixerKwargs.generator] = generator
 
         for mixer in self.mixers.values():


fast_llm/layers/decoder/stochastic_mixer.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def preprocess(self, batch: torch.Tensor, kwargs: dict[str, typing.Any]) -> None:
+        """
+        Preprocess for all mixers.
+
+        Since we don't know which mixer will be selected during training,
+        we need to preprocess for all of them. This includes things like
+        attention masks, rotary embeddings, etc.
+        """
+        for mixer in self.mixers:
```
jlamypoirier: There could be name conflicts. Consider namespace?

fast_llm/models/gpt/conversion/apriel.py
```diff
@@ -232,16 +232,81 @@ class AprielMamba2BlockConverter(MistralBlockConverter):
...
+        Assert.custom(isinstance, config, StochasticMixerConfig)
+        inference_mixer = config.mixers[config.main_mixer_index]
+        mixer_type = type(inference_mixer)
+        converter_class = cls._mixer_block_converters.get(mixer_type)
+        if converter_class is None:
+            raise NotImplementedError(f"No converter for mixer type: {mixer_type.__name__}")
+        return converter_class.mixer_converter_class.export_config(inference_mixer)
+
+    @classmethod
+    def get_converters(
```
jlamypoirier: How about import? I don't think this will work.

jlamypoirier (commented):
Looks good, some minor comments

fast_llm/layers/decoder/config.py
```diff
@@ -71,6 +80,72 @@ def _from_dict(cls, default: dict[str, typing.Any], strict: bool = True) -> typi
...
+    """
+    Stochastic mixer that uniformly samples from multiple mixer options during training.
+
+    For supernet training, each forward pass randomly selects one mixer to execute,
+    training all mixers with different subsets of data.
+    """
+
+    _abstract = False
+
+    mixers: list[MixerConfig] = Field(
```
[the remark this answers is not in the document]
tscholak: now using `dict[str, MixerConfig]`

fast_llm/models/gpt/conversion/apriel.py
tscholak: import doesn't work as usual for stochastic mixers. we use `drop_on_export=True` for non-main mixers, and HF checkpoints only contain the main mixer. I think the correct way to handle this is to either support stochastic mixers in hf (out of scope) or initialize all other mixers randomly while importing only the main mixer.

fast_llm/models/gpt/conversion/apriel.py
```diff
@@ -313,20 +379,32 @@ def export_config(cls, config: BlockSequenceConfig) -> dict:
...
         converters = []
-        for block_index in range(config.num_blocks):
-            block_config = config.blocks[config.pattern[block_index % len(config.pattern)]]
-            converters += cls.block_converter_class.get_converters(
-                block_config,
-                f"{fast_llm_prefix}.{block_index}",
-                f"{hf_prefix}.{block_index}",
-                drop_on_export,
-            )
+        match config:
```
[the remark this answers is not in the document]
tscholak: uses if-else with instance type check now.

fast_llm/engine/checkpoint/huggingface.py
```diff
@@ -73,10 +73,38 @@ def _serialize_metadata(self, config: CheckpointSaveMetadataConfig, metadata: Ch
             "format": "pt",
         }
 
+    def _initialize_missing_parameters(self) -> None:
+        # Parameters that exist in the model but not in the checkpoint import converters
+        missing_params = set(self._export_converters.keys()) - {
+            weight_converter.fast_llm_name[0]
```
jlamypoirier: There can be multiple fast_llm names.
```
{ name for weight_converter in self._import_converters.values() for name in weight_converter.fast_llm_name}
```

fast_llm/engine/checkpoint/huggingface.py
```diff
@@ -73,10 +73,38 @@ def _serialize_metadata(self, config: CheckpointSaveMetadataConfig, metadata: Ch
...
+            for weight_converter in self._import_converters.values()
+            if weight_converter.fast_llm_name
+        }
+
+        print(f"[INIT DEBUG] Checking for missing parameters in HuggingFace checkpoint...")
```
jlamypoirier: Please use `logger.info` to match other logs (and no need for `[INIT DEBUG]`)

fast_llm/layers/decoder/config.py
```diff
@@ -55,6 +63,13 @@ def _from_dict(cls, default: dict[str, typing.Any], strict: bool = True) -> typi
         return super()._from_dict(default, strict=strict)
 
 
+class SamplingStrategy(str, enum.Enum):
```
jlamypoirier: `StochasticMixerSamplingStrategy`

fast_llm/layers/decoder/stochastic_mixer.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns:
+            Name of the mixer to use for this forward pass.
+        """
+        if not self.training:
+            # Use main mixer for inference
+            return self._config.main_mixer_name
+
+        # Sample index in training mode
+        generator = self._distributed.tp_generator if self._sequence_parallel else self._distributed.pp_generator
```
jlamypoirier: I don't think that's right, `tp_generator` will result in different tensor ranks selecting different mixers.

tests/models/test_checkpoint.py
```diff
@@ -210,6 +219,36 @@ def _compare_architectures(config_ref: FastLLMModelConfig, config_test: FastLLMM
...
+    return model_config_class.from_dict(config_dict)
+
+
+def _load_config_from_checkpoint(checkpoint_path: pathlib.Path, model_config_class) -> FastLLMModelConfig:
+    """Load model config from checkpoint metadata.yaml."""
+    config_dict = yaml.safe_load(checkpoint_path.joinpath("metadata.yaml").open("r"))["config"]
+    return model_config_class.from_dict(config_dict)
+
+
+def _is_lossy_hf_conversion(checkpoint_format: type[CheckpointFormat] | None, base_model_config) -> bool:
```
jlamypoirier: How about an explicit marker in `ModelTestingConfig` instead?

tscholak:
Hi @jlamypoirier! Thanks for the thorough review! I've addressed all your feedback and cleaned up the
  PR significantly. Here's what changed since your last review:

Stochastic mixer sampling fix (your main concern):

- Added iteration to BlockKwargs and propagated it through preprocess_batch → kwargs
- Create a torch CPU generator in preprocess(), seeded with iteration
- Sample mixers in forward() using torch.multinomial with the CPU generator
- This ensures all DP/TP/PP ranks sample the same mixer sequence per batch, while different layers
  sample different mixers deterministically
- Removed the TP/PP generator usage that was causing rank inconsistencies
- Store sampling probabilities on CPU to avoid device transfers

Code cleanup:

- Removed all debug print statements and irrelevant changes
- Removed stochastic mixer support from old apriel converter (only supported in apriel2 now)
- Removed tools/supernet_beam_search.py (will implement differently later)
- Renamed SamplingStrategy → StochasticMixerSamplingStrategy
- Use normalize_probabilities for sampling weight validation (consistent with dataset blending)

Note: I added a new apriel2 conversion module for the newer model format. This doesn't need review as it's separate from the stochastic mixer changes. Apriel hybrid SSM export is not supported. Stochastic layers are only in apriel2.

Ready for another look!

fast_llm/layers/decoder/stochastic_mixer.py
```diff
@@ -0,0 +1,167 @@
...
+        else:
+            raise NotImplementedError(f"Sampling strategy {self._config.sampling_strategy} not implemented")
+
+        logger.info(
+            f"Initialized StochasticMixer with {len(self.mixers)} mixers: "
+            f"{', '.join(f'{name}={type(mixer).__name__}' for name, mixer in self.mixers.items())} "
+            f"(main={self._config.main_mixer_name})"
+        )
+
+        # Mark all mixer parameters with allow_no_grad since only one mixer
```
jlamypoirier: One thing to keep in mind, all mixers will still go through gradient reduction and optimization step, the unused ones will get a zero gradient. Probably still ok, but it's suboptimal and could affect optimizer momenta. I don't think there is an easy fix though.

PR #377, opened 2025-10-15
Dataset interface

jlamypoirier:
# ✨ Description

Part of the data rework:
* Drop the generic / gpt specialization of indexed dataset types. Instead, use generics to specify a sample type.
* Add `Sample` and `Batch` constructs (placeholders for now)
* Remove the tokenizer from `GPTData`, since it's rarely used and not by the data itself. Instead, use separate tokenizers where needed (Fim, Preparator (already present), lm eval).
* Start moving away from numpy in favor of torch so things are more uniform (ex. `Sample` and `Batch` both use torch tensors.)
* Other minor tweaks.

Note: since this is part of a bigger set of changes, it does contain changes that don't immediately make sense but will be useful layer, as well as messy temporary solutions. (See #376 for more info on where this is going).


tscholak:
Great to carve this out!

I'm wondering how to stage these changes. Switching to the new binary format is a breaking change, and we'd need to reprocess all currently used training data. This needs to be properly timed and announced. Do we have backwards compatibility for already processed data?

tscholak (approved):
LGTM!

PR #378, opened 2025-10-16
Language model sample

jlamypoirier:
# ✨ Description

* Replace `GPTSample` and `GPTBatch` with `LanguageModelSample` and `LanguageModelBatch`, encapsulating much of the related functionality and simplifying much of the code that use them (ex. `SampledIndexedDataset`, `GPTBaseModel.preprocess_batch`, `gpt_data_collate_fn`
* Make `SampledIndexedDataset` agnostic of the sample type.
* Change spans to use a standard range format, i.e. (first, last + 1) instead of (first, last). (Spans are still stored in (first, last) format in the binary file, this will be changed for the new binary format.)
* Redo much of the code for preference spans. `SampledIndexedDataset` was using an entirely different code path for preference spans avoiding multi-document samples (for historical reasons I think?), I dropped it and made it use the common code path. This does mean a change in behavior (ex. multi-document samples), but that's an improvement. I have some doubts about the dpo ploss function though, I suspect the log softmax needs to be calculated separately for each document.
* Datasets now always provide sequence lengths, and move `cross_document_attention` to `AttentionConfig` (from `BatchConfig`) so the attention layer decides itself whether to use varlen or not. Replace `use_flash_attention` with a more generic `implementation` enum. (See discussion in #370.) Add separate `LanguageModelEmbeddingsConfig.cross_document_position_embeddings` since absolute position embeddings may also use the sequence lengths.


tscholak (approved):
this looks ok to me. I think we should merge.

For others looking at this, the most important changes are in:

* [‎fast_llm/data/dataset/sampled.py‎](https://github.com/ServiceNow/Fast-LLM/pull/378/files#diff-ccb5922b58f4ad036a3adc0c009a9480cf92afb66b08a44b9dbfe873cb664d20)
* [‎fast_llm/data/sample/language_model.py‎](https://github.com/ServiceNow/Fast-LLM/pull/378/files#diff-9dafec68febc1f36396df5fa75747ea2fcbd0dc1bb1a75769ee3a0e8ff8cd4de)
* [‎fast_llm/data/sample/range.py‎](https://github.com/ServiceNow/Fast-LLM/pull/378/files#diff-73409d043693d4e793386272603a62d368aba09683e9522a141d660e69a3e295)
* [‎fast_llm/data/sample/token.py‎](https://github.com/ServiceNow/Fast-LLM/pull/378/files#diff-606cf2792cf9d5ecd5c41880af2f5ecd10b8220b0afb97245d79431f5f1eb38a)
* [‎fast_llm/functional/dpo.py‎](https://github.com/ServiceNow/Fast-LLM/pull/378/files#diff-31111ff01831d6540e1d81146f23a6ec805e01733dce0acb9f0cb4cae1106d6d)
* [‎fast_llm/models/gpt/model.py‎](https://github.com/ServiceNow/Fast-LLM/pull/378/files#diff-ad9bedaf18682bf87a3deaaacf1af4b293c130c1d89336b05143baa61bff9a37)
* [‎tests/functional/test_functional.py‎](https://github.com/ServiceNow/Fast-LLM/pull/378/files#diff-2d47aecd6e530374b518d33afdcd7c8ff9fa5adee2c6bc43b49fed934068a76e)

a few things that we should definitely make sure of, ideally via tests (if we haven't already):

* span cropping, offsetting, and truncation arithmetic is correct
* loss masking masks correctly
* dpo works correctly

PR #397, opened 2025-11-27
Apriel 2

tscholak:
This PR enables **declarative architecture surgery** on Apriel2 models. You can now convert a standard attention-based model into exotic architectures (stochastic supernets, hybrid Mamba/GDN models) with a single command and a YAML config.

To convert Apriel-1.5-15b-Thinker into a stochastic supernet with 3 mixer types per layer:

```bash
python convert.py ServiceNow-AI/Apriel-1.5-15b-Thinker /tmp/apriel2-supernet \
    --surgery examples/stochastic_supernet.yaml --verbose
```

The converter shows you exactly what it's doing with a visual plan tree:

```
============================================================
CONVERSION PLAN
============================================================
├── model/
│   ├── [embed_tokens, norm]/
│   │   └── weight ← language_model.model.[embed_tokens, norm].weight
│   ├── vision_encoder/
│   │   └── encoder/blocks/[0..23]/...
│   └── decoder/
│       └── blocks/
│           └── [0..47]/
│               ├── mixer/
│               │   └── mixers/
│               │       ├── attention/
│               │       │   └── self_attn/[q,k,v,o]_proj/weight ← ...self_attn.[q,k,v,o]_proj.weight
│               │       ├── sliding_window/
│               │       │   └── self_attn/[q,k,v,o]_proj/weight ← ...self_attn.[q,k,v,o]_proj.weight
│               │       └── gated_delta_net/
│               │           └── gdn/
│               │               ├── in_proj_qkvz/weight = [Q[:160,:]; K[:160,:]; V slices...; 𝟎(640×5120)]
│               │               ├── in_proj_ba/weight = 𝟎(64×5120)
│               │               ├── out_proj/weight ← ...self_attn.o_proj.weight
│               │               ├── conv1d/weight = scaled_identity_conv(7680×1×4)
│               │               ├── A_log = A_log(32)
│               │               └── norm/weight = 𝟏(160)
│               ├── mlp/[gate,up,down]_proj/weight ← ...
│               └── [input_layernorm, post_attention_layernorm]/weight ← ...
└── lm_head/weight ← language_model.lm_head.weight
============================================================
Converting: 100%|████████████████████████████████| 1185/1185 [00:40<00:00]
Conversion complete! Output saved to /tmp/apriel2-supernet
```

The result is a sharded Apriel2 checkpoint with attention + sliding window + gated delta net mixers in every layer.

The surgery config for this looks like this:

```yaml
decoder:
  type: fixed
  block:
    mixer:
[...]

PR #399, opened 2025-12-02
llava conversion fixes

RaymondLi0:
# ✨ Description

Small changes to be able to load [Apriel-1.5-15B](https://huggingface.co/ServiceNow-AI/Apriel-1.5-15b-Thinker):
- remove `projector_intermediate_size` from llava_hybrid and llava converter (use text-config's hidden-size, like in llava)
- map hf's `gelu` activation
- fix some of the hf weight prefixes

With these changes, I am able to load Apriel-1.5-15B in fast-llm.
This also points to the issue that these were not caught by the conversion tests.


fast_llm/functional/config.py
```diff
@@ -39,6 +39,7 @@ class ActivationType(enum.StrEnum):
     An enum for the available activation types for the MLP layer.
     """
 
+    gelu_gaussian = "gelu_gaussian"
```
jlamypoirier: There is no point in having a separate gelu because we're not going for exact numerical match, and we already have bigger numerical differences anyway. HF gelu and gelu_pytorch_tanh should both map to our standard gelu

fast_llm/functional/config.py
```diff
@@ -84,7 +84,15 @@ def _set_activation_fn_map() -> None:
     ActivationType.squared_relu: "relu2",
     ActivationType.identity: "identity",
 }
-_ACTIVATION_HF_NAMES_INV = {value: key for key, value in _ACTIVATION_HF_NAMES.items()}
+# gelu and gelu_pytorch_tanh both map to our standard gelu
```
jlamypoirier: Please use
https://github.com/ServiceNow/Fast-LLM/pull/397/files#diff-44a40ea2ff12b8fae0614252d0e9aa8b860e826fb9631d0d309393d1fb5ebfdcR88

fast_llm/models/multimodal/conversion/llava.py
```diff
@@ -184,8 +184,6 @@ def export_config(cls, config: MLPConfig) -> dict:
         return {
             "projector_hidden_act": config.activation.hf_name,
             "multimodal_projector_bias": config.add_linear_biases,
-            # Not in LlavaConfig, but needed for consistency check in LlavaBaseModelConverter.
```
jlamypoirier: Why removing? This is essential to ensure compatibility.

fast_llm/models/multimodal/conversion/llava.py
```diff
@@ -245,13 +243,13 @@ def export_config(cls, config: VisionEncoderConfig) -> dict:
     def get_converters(cls, config: VisionEncoderConfig) -> list[WeightConverter]:
         return [
             *cls.embeddings_converter_class.get_converters(
-                config.embeddings, "vision_encoder.embeddings", "model.vision_tower"
+                config.embeddings, "vision_encoder.embeddings", "vision_tower"
```
jlamypoirier: Why these changes? The current names are required for `LlavaForConditionalGeneration` and confirmed to work. The `model` prefix is explicitly needed for `LlavaForConditionalGeneration` `https://github.com/huggingface/transformers/blob/main/src/transformers/models/llava/modeling_llava.py#L316`and the language model is a `MistralModel` which takes no `model` prefix.

RaymondLi0: Hmm indeed, it's strange.
Without all these changes, we're not able to load https://huggingface.co/ServiceNow-AI/Apriel-1.5-15b-Thinker/tree/main in fast-llm. The weights in that model somehow match this different format with `language_model.model...`

fast_llm/models/multimodal/conversion/llava.py
jlamypoirier: The projector intermediate size needs to match with the LM hidden size, which is not guaranteed on the Fast-LLM size. The entry is not in the final output, it's there specifically for the assertion in https://github.com/ServiceNow/Fast-LLM/pull/399/files#diff-319643f77a4055995eb8f844aee095266ba3b15fa11f52e16acd89386058e51bL314. A failing assertion points to an actual error elsewhere. 

What do you mean by "load Apriel-1.5"? Shouldn't that go through import?

fast_llm/models/multimodal/conversion/llava.py
jlamypoirier: My understanding is that there are two equivalent ways to see the model. It can either be a `LlavaForConditionalGeneration` with a `MistralModel` text model, or a `LlavaModel` with a `MistralForCausalLM`.  Main exports  in the first format, but the dev branch seems to use the second one, though is still uses `LlavaForConditionalGeneration` as the architecture (maybe `_checkpoint_conversion_mapping` addresses the mismatch?)

I'd think the first option is more appropriate, but I could be wrong. Maybe we could just support both cases.

fast_llm/models/multimodal/conversion/llava.py
oleksost: `The projector intermediate size needs to match with the LM hidden size` -- @jlamypoirier why should the projector's hidden size depend have anything to do with model's hidden size? IIUC intermediate size fo the projector can be anything, its the projector's output that must match?

oleksost:
Hey @RaymondLi0 @jlamypoirier can this be merged soon? These changes (or at least some of these) are required to work with Apriel 1.6/1.5 currently as it seems

PR #401, opened 2025-12-04
Fix and test rotary 2d

jlamypoirier:
# ✨ Description

Frequencies were slightly different from Pixtral. Also added a test, which is a simplified version of the one in https://github.com/ServiceNow/Fast-LLM/pull/397/files#diff-9d27d91b42178626702d41f5a438bc49ed3e03ce54bb75c55322c30c0dfeccdfR1


tscholak (approved):
LGTM! thanks!

PR #419, opened 2025-12-12
Remove mamba and discrete mamba 2, rename mamba 2 to mamba

jlamypoirier:
As discussed. I didn't remove them from the external models.


tscholak:
`causal_conv1d_cuda` can't be found, on cpu-only CI. something is off with the cuda-gating for the ssm tests

PR #420, opened 2025-12-12
Add metadata to dataset config files

jlamypoirier:
# ✨ Description

Gather all available metadata from readers during dataset preparation and add it to the yaml config file. Works with both blending and splitting. Example:
```
$ cat /tmp/fast_llm_tests/common/dataset/dataset_with_image_patches_55_132/fast_llm_config.yaml 
config:
  path: shard_0_0.fast_llm_dataset
  type: memmap
metadata:
  image_patches:
    data_type: uint8
    num_documents: 1000
    num_patch_groups: 1355
    num_patches: 10707
    num_pixels: 513936
    patch_shape:
    - 3
    - 4
    - 4
  num_tokens: 59145
  tokens:
    data_type: int32
    num_documents: 1000
    num_tokens: 59145
```

The file format is different from before (config is now inside of a dict instead of at top level), but the old format can still be loaded.

### DIFF
--- a/fast_llm/data/sample/range.py
+++ b/fast_llm/data/sample/range.py
@@ -33,12 +33,15 @@ class RangeSample(Sample):
 
     @classmethod
     def from_documents(cls, documents: typing.Iterable[typing.Self]) -> typing.Self:
+        """
+        Used to merge ranges from multiple documents, i.e. when multiple docuemnts are packed together.
+        """
         document: RangeSample
         ranges = []
         sample_size = 0
         for document in documents:
             for begin, end in document.ranges:
-                ranges.extend((begin + sample_size, end + sample_size))
+                ranges.append((begin + sample_size, end + sample_size))
             sample_size += document.sample_size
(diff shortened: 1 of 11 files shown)


fast_llm/data/dataset/gpt/config.py
```diff
@@ -64,7 +64,12 @@ def build(self, preprocessing: PreprocessingConfig) -> SamplableDataset[SampleTy
 
     def _load_config(self) -> SampledDatasetConfig[SampleType]:
         assert self.path.is_file(), f"File {self.path} does not exist."
-        return SampledDatasetConfig[SampleType].from_dict(self._convert_paths(yaml.safe_load(self.path.open("r"))))
+        config = yaml.safe_load(self.path.open("r"))
+        Assert.eq(config.keys(), {"config", "metadata"})
```
tscholak: @jlamypoirier this is causing crashes now. remove?
```suggestion
```

PR #424, opened 2025-12-15
Add support for gloo

jlamypoirier:
# ✨ Description

Make Fast-LLM work with the gloo distributed backend. Context: I was hoping it would fix some random NCCL crashes/timeouts in distributed tests with many processes, allowing us to run more tests in parallel (and without the need for multiple physical). However, the problem still happens with gloo, so keeping the test in nccl for now.


tscholak (approved):
LGTM!

PR #428, opened 2025-12-17
[Pipeline RL] Add support for PipelineRL

jlamypoirier:
This PR provides the initial integration with **PipelineRL** with GRPO loss.

It introduces:

- **Streaming Redis Dataset** — capable of consuming documents such as rollouts from an external Redis stream.
- **Trainer Callback System** — supports callbacks for events like `training_started`, `step_finished`, and `training_finished`.
- **Redis-based Callback Implementation with Weights Broadcast Mechanism** — uses a separate external NCCL rendezvous point to broadcast updated model weights in real time to inference servers, and a Redis stream to broadcast training events.

This enables seamless coordination between Fast-LLM training and PipelineRL-based inference or orchestration components.

### DIFF
--- a/fast_llm/data/dataset/streaming.py
+++ b/fast_llm/data/dataset/streaming.py
@@ -156,11 +156,11 @@ class RedisStreamingDataset[ConfigType: StreamingDatasetConfig, DocumentType: La
                 noack=True,
             )
             if messages:
-                start_time = time.time()
                 for stream_key, messages_ in messages:
                     assert stream_key == REDIS_DATA_STREAM.encode()
                     for message_id, message in messages_:
                         yield RedisStreamingDocumentData.from_message(message).to_document()
+                start_time = time.time()
 
-            if (t := time.time() - start_time) > self._config.timeout:
+            elif (t := time.time() - start_time) > self._config.timeout:
                 raise TimeoutError(f"No document received after {t} seconds")
--- a/fast_llm/data/document/language_model.py
+++ b/fast_llm/data/document/language_model.py
@@ -161,12 +161,13 @@ class LanguageModelBatch(TokenBatch):
                 length_cumsum = torch.tensor([0] + cropped_lengths, device=self.device).cumsum(0)
                 label_count_cumsum = mask_cumsum[length_cumsum]
                 labels_per_document = label_count_cumsum[1:] - label_count_cumsum[:-1]
-                # Expand to one entry per token: find each token's document via the sorted
+                # Expand to one entry per token: find each token's document index via the sorted
                 # length cumsum, then look up that document's label count.
-                doc_per_token = torch.searchsorted(
+                # TODO: Document index already computed in `LengthModelInputPreprocessor`.
+                document_index = torch.searchsorted(
                     length_cumsum[1:], torch.arange(len(mask), device=self.device), side="right"
                 )
-                label_counts = labels_per_document[doc_per_token][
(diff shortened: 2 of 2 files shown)


fast_llm/data/dataset/streaming.py
```diff
@@ -0,0 +1,111 @@
...
+                # No explicit ACK: messages are processed immediately; on rank failure the job restarts,
+                # so message loss is acceptable and simplifies coordination
+                noack=True,
+            )
+            if messages:
+                for stream_key, msgs in messages:
+                    assert stream_key == REDIS_DATA_STREAM.encode()
+                    for msg_id, msg_data in msgs:
+                        processed += 1
+                        # TODO: or do it after processing all received messaged then count > 1?
```
bigximik: We do not need this for now if we stick with consumer groups only. The producer can rely on the group `lag` to control the production rate and on `last-delivered-id` to safely trim the stream.

fast_llm/data/dataset/config.py
```diff
@@ -298,3 +299,56 @@ def build(self, preprocessing: PreprocessingConfig) -> "IndexedDataset[SampleTyp
...
+
+@config_class(dynamic_type={SampledDatasetConfig: "streaming"})
+class StreamingDatasetConfig[SampleType: LanguageModelSample](RedisConfig, SamplableDatasetConfig[SampleType]):
+    """
+    Configuration for a streaming dataset that reads training data from a Redis stream.
+    """
+
+    _abstract = False
+
+    acknowledge_interval: int = Field(
```
bigximik: This is also not needed if we only use consumer gorups, see above comment for implementation.

fast_llm/engine/training/streaming.py
```diff
@@ -0,0 +1,77 @@
...
+        if self._do_broadcast:
+            self._client = self._config.get_client()
+            init_method = f"tcp://{config.broadcast.host}:{config.broadcast.port}"
+            logger.info(f"Waiting for weights broadcast rendezvous at {init_method} ...")
+            # TODO: Create a custom process group instead.
+            self._process_group = torch.distributed.init_process_group(
+                backend="nccl",
+                init_method=init_method,
+                world_size=config.broadcast.external_world_size + 1,
+                rank=0,
```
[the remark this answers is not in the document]
jlamypoirier: We also have control over the external system, we can make it so. It's easier to sync program with hard-coded values than syncing config files.

fast_llm/layers/language_model/loss/grpo.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+    per_sample_loss = per_sample_loss * loss_mask
+    loss = per_sample_loss.mean()
+
+    # Sum of per-sequence mean log-probs, matching pipelinerl's new_logprobs metric:
+    #   sum_sum(new_logprobs / num_labels_in_seq, masks_shifted, segments)
+    # Dividing by num_labels_in_seq (span length broadcast per token) and summing over masked
+    # tokens gives mean logprob per sequence; summing those across sequences matches the deepspeed
+    # convention exactly (segments are redundant once num_labels_in_seq is correct).
+    new_logprobs_mean = None if num_labels_in_seq is None else (new_log_probs * loss_mask / num_labels_in_seq).sum()
```
rafapi: here we should do `num_labels_in_seq.clamp(min=1.0)` to avoid division by 0 in cases where complete document is masked

rafapi (dismissed):

PR #432, opened 2025-12-21
Add IS evaluator

tscholak:


jlamypoirier:
**Claude Opus 4.8 triage note:**

Status: draft, conflicting with `main` and well behind it, original author no longer active.

The branch is a multi-feature stack. The headline feature is a forward-KL / importance-sampling evaluator (IS accuracy + ESS for teacher-trace / distillation evaluation) under `fast_llm/engine/evaluation/forward_kl/`. It's bundled with several unrelated changes — an Apriel2 cache refactor, a Qwen2 converter, SFT conversation-format data prep, per-layer bias support — and some of those appear to overlap work that has since landed independently on `main` (would need checking before reuse).

What's missing to move forward:
1. A decision on whether the IS / forward-KL teacher-trace evaluator is still wanted.
2. If so, the evaluator needs to be extracted from the surrounding bundle and re-implemented on current `main` (the rest is either stale or already merged elsewhere).

The evaluator is the only unique salvageable piece. Leaving open pending that decision.

PR #460, opened 2026-01-26
Fix default lm loss

jlamypoirier:
# ✨ Description

Avoid adding the default loss at the config level as it could be incorrectly kept in case of config overriding.


oleksost (approved):

oleksost:
Not sure why those tests fail, they don't fail locally for me. Maybe related to some package versions

jlamypoirier:
> Not sure why those tests fail, they don't fail locally for me. Maybe related to some package versions

It's clearly not related to this PR, merging anyway. Most likely has to do with transformers 5.0.0 which was released today

PR #461, opened 2026-01-27
Bump versions, restrict transformers

jlamypoirier:
# ✨ Description

* Bump minimum versions (taken from built image)
* Restrict transformers version until we support v5 #462
* Rerun and fix code formatting for the whole repo using the new versions.


tscholak (approved):
LGTM

PR #426, opened 2025-12-16
[Prototype] Normalising by valid tokens

oleksost:
# ✨ Description
Fixes several issues related to loss masking, added loss matching checks in the model tests.

#### 1. Cross entropy loss seem to be incorrectly averaged when loss masks are used?
**Symptom:**
- no symptom really, since the reference torch implementation was averaging incorrectly [here](https://github.com/ServiceNow/Fast-LLM/blob/d2018f92ab79950da2b72b7be482f9225496acf7/fast_llm/functional/cross_entropy.py#L43)
- was discovered because the `df4` tests with loss masking were **not** failing for CE loss (they did for rev. KL loss)

**Proposed solution:**
- IIUC when computing loss and grad mean we need to divide by the number of valid tokens (`loss_mask.sum()`) when mask is used and not by all tokens (similar as in rev. KL case)

#### 2.Incorrect grad & loss averaging with loss masking & grad. accumulation

 When training with micro-batches grad. accumulation & loss masking, loss and grads are incorrectly averaged across   micro-batches because of different number of valid tokens in each micro-batch (i.e. loss_mask.sum() are different across micro-batches). 

**Symptom:**
- test `tests/models/test_model.py::test_and_compare_model[mistral_reverse_kl-df4]@dependency_group_15` is failing

**Proposed solution:** 
- calculate global number of valid tokens (requires an additional iterations over micro-batches in the runner) --> scale `grad_output` by `loss_mask.sum() / total_valid_tokens` to make sure the average over micro-batches is correct.
- added loss comparison to the model tests: bot `pytest tests/models/test_model.py --models mistral_reverse_kl` and `pytest tests/models/test_model.py --models mistral_distill_logits` pass now, the losses and grads seem to match the `simple` baseline.

## 🔍 Type of change

Select all that apply:

- [x] 🐛 **Bug fix** (non-breaking change that addresses a specific issue)
- [ ] 🚀 **New feature** (non-breaking change that adds functionality)
- [ ] ⚠️ **Breaking change** (a change that could affect existing functionality)
- [ ] 📈 **Performance improvement/optimization** (improves speed, memory usage, or efficiency)
- [ ] 🛠️ **Code refactor** (non-functional changes that improve code readability, structure, etc.)
- [ ] 📦 **Dependency bump** (updates dependencies, including Dockerfile or package changes)
- [ ] 📝 **Documentation change** (updates documentation, including new content or typo fixes)
[...]

PR #476, opened 2026-03-21
[WIP][PipelineRL] Normalization of new_logprobs and addition of other RL metrics

bigximik:
Add generic denominator_batch_field to LossDef so any metric can be normalized by a pre-computed per-micro-batch scalar from the batch context, bypassing TP/SP/PP splitting entirely.

For grpo_new_logprobs: compute num_docs = (labels_per_document > 0).sum() in language_model.py before any parallel splitting, giving a true per-document average regardless of variable document lengths.

Only sequence_data_rank==0 contributes to num_docs. The runner all_reduces the denominator across the data group (which includes SDP ranks); if every SDP rank reported its own num_docs, a single document processed by SDP=2 would be counted twice, halving the metric.

Also clamp num_labels_in_seq to avoid 0/0=nan for padding segments or fully-masked documents (loss_mask=0 there so the numerator is 0 too).

Tests verify:
- num_docs counts only unmasked documents
- padding segments (pad_to_size) are excluded
- with SDP=2, only rank 0 contributes num_docs so the all_reduce SUM across SDP ranks gives the correct denominator

# ✨ Description

Please provide a brief summary of the changes, relevant motivation, and context.
Include any related issue numbers or links to discussions, and explain why this change is necessary.

Closes # 

## 🔍 Type of change

Select all that apply:

## 📝 Changes

1. Change A
2. Change B

## ✅ Checklist

### General

- [ ] 📜 I have read and followed the [contributing guidelines](https://servicenow.github.io/Fast-LLM/contributing/contributing).

### Testing

### Performance Impact

---

## 🗒️ Additional Notes

### DIFF
--- a/fast_llm/data/document/language_model.py
+++ b/fast_llm/data/document/language_model.py
@@ -39,6 +39,9 @@ class LanguageModelTargetInput(ModelInput):
 class LanguageModelInput(TokenModelInput):
     targets: list[LanguageModelTargetInput] = dataclasses.field(default_factory=list)
     image_patches: PatchModelInput | None = None
+    # Number of documents with at least one response token in this micro-batch.
+    # Computed before any TP/SP/PP splitting; used to normalize per-document metrics.
+    num_docs: int | None = None
 
     def set_children_attributes(self) -> None:
         if self.image_patches is not None:
@@ -58,6 +61,7 @@ class LanguageModelInput(TokenModelInput):
             LanguageModelKwargs.advantages: [target.advantages for target in self.targets],
             LanguageModelKwargs.old_log_probabilities: [target.old_log_probabilities for target in self.targets],
             LanguageModelKwargs.label_counts: [target.label_counts for target in self.targets],
+            LanguageModelKwargs.num_docs: self.num_docs,
         }
         if self.image_patches is not None:
             out.update(self.image_patches.to_kwargs())
@@ -161,6 +165,17 @@ class LanguageModelBatch(TokenBatch):
                 length_cumsum = torch.tensor([0] + cropped_lengths, device=self.device).cumsum(0)
                 label_count_cumsum = mask_cumsum[length_cumsum]
                 labels_per_document = label_count_cumsum[1:] - label_count_cumsum[:-1]
+                # Track documents with at least one response token for per-document metric
+                # normalization. Only counted on sequence_data_rank==0 so that when the runner
+                # all_reduces the denominator across the data group (which includes SDP ranks),
+                # each document is counted exactly once even though all SDP ranks see the same
+                # documents (but process different token slices of them).
+                if model_input.num_docs is None:
+                    model_input.num_docs = (
+                        int((labels_per_document > 0).sum().item())
+                        if config.distributed.sequence_data_rank == 0
+                        else 0
+                    )
                 # Expand to one entry per token: find each token's document index via the sorted
                 # length cumsum, then look up that document's label count.
                 # TODO: Document index already computed in `LengthModelInputPreprocessor`.
--- a/fast_llm/data/document/language_model.py
+++ b/fast_llm/data/document/language_model.py
@@ -177,11 +177,15 @@ class LanguageModelBatch(TokenBatch):
                 # each document is counted exactly once even though all SDP ranks see the same
                 # documents (but process different token slices of them).
                 if model_input.num_docs is None:
-                    model_input.num_docs = (
-                        int((labels_per_document > 0).sum().item())
-                        if config.distributed.sequence_data_rank == 0
-                        else 0
-                    )
+                    # Skip .item() on meta tensors (called during setup/shape inference).
+                    if labels_per_document.is_meta:
+                        model_input.num_docs = 0
+                    else:
+                        model_input.num_docs = (
+                            int((labels_per_document > 0).sum().item())
+                            if config.distributed.sequence_data_rank == 0
+                            else 0
+                        )
                 # Expand to one entry per token: find each token's document index via the sorted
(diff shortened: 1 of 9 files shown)


fast_llm/data/document/language_model.py
```diff
@@ -39,6 +39,9 @@ class LanguageModelTargetInput(ModelInput):
 class LanguageModelInput(TokenModelInput):
     targets: list[LanguageModelTargetInput] = dataclasses.field(default_factory=list)
     image_patches: PatchModelInput | None = None
+    # Number of documents with at least one response token in this micro-batch.
```
jlamypoirier, line 42: Why not just using the plain document count? Is there any scenario where we would expect response-free documents and need to take them into account? Seems like they shouldn't be in the batch to begin with because they don't contribute to the loss...

fast_llm/engine/base_model/config.py
```diff
@@ -109,6 +109,9 @@ class LossDef:
     name: str
     formatted_name: str
     # The number of times this loss is evaluated by the model for each micro-batch. Used as a denominator for averaging.
-    # TODO: Allow variable count?  Would need a reduction across PP devices.
     count: int = 1
     dtype: DataType = DataType.float32
+    # If set, normalize this metric by summing values from context.batch[i][denominator_batch_field]
+    # across micro-batches and DP ranks, instead of using count * data_parallel * num_inputs.
+    # The field must be a scalar (int or float) pre-computed before any TP/SP/PP splitting.
+    denominator_batch_field: str | None = None
```
jlamypoirier, line 117: Somewhat redundant with `count`. Forcing the denominator to be in a scalar-valued kwarg field may not be generic enough, I'd suggest returning losses as a `(numerator, denominator)` tuple instead, or maybe reducing the denominator beforehand?

fast_llm/data/document/language_model.py
```diff
@@ -161,6 +171,21 @@ def _get_model_input(
...
+                # normalization. Only counted on sequence_data_rank==0 so that when the runner
+                # all_reduces the denominator across the data group (which includes SDP ranks),
+                # each document is counted exactly once even though all SDP ranks see the same
+                # documents (but process different token slices of them).
+                if model_input.num_docs is None:
+                    # Skip .item() on meta tensors (called during setup/shape inference).
+                    if labels_per_document.is_meta:
+                        model_input.num_docs = 0
+                    else:
+                        model_input.num_docs = (
```
jlamypoirier, line 184: We could just reduce over the batch data group instead?

fast_llm/engine/schedule/runner.py
```diff
@@ -287,20 +287,39 @@ def run_step(
...
                 if losses:
-                    loss_count = (
-                        self._loss_definitions[name].count
-                        * self._distributed_config.data_parallel
-                        * context.schedule.config.num_inputs
-                    )
-                    reduced_loss = torch.stack(losses).sum() / loss_count
-                    if self._distributed.data_group:
-                        all_reduce(reduced_loss, group=self._distributed.data_group)
+                    denom_field = loss_def.denominator_batch_field
```
jlamypoirier, line 294: This works for metrics, but not for actual losses used for training which would need to reproduce this code. I think this could be addressed by reducing the values at the beginning of the batch, ex. in `ModelInput.reduce_counts` so we can use the counts freely from anywhere.

tests/data/test_preprocessing.py
```diff
@@ -53,3 +55,163 @@ def test_preprocessing(tokens, loss_masking_spans):
 
     Assert.eq(len(model_input.targets), 1)
     Assert.all_equal(model_input.targets[0].tokens, torch.cat(label_tokens)[1:])
+
```
jlamypoirier, line 58: I think most of these can be combined into a single parameterized test?

jlamypoirier (commented):
Looks good, but I would like to look at the problem in a more generic context, especially since we've had multiple requests to address normalization for all losses and metrics (ex. #426).

fast_llm/data/document/language_model.py
bigximik: Agree that response-free documents shouldn't be in the batch. However the > 0 check is needed because of padding — by the time we reach _get_model_input, padding is just another entry in lengths and structurally indistinguishable from a real document. To get the true document count cleanly, we can store num_real_docs = len(documents) in from_documents before
   padding is appended.

fast_llm/engine/base_model/config.py
bigximik: The reason of not returning (numerator, denominator) from the loss is that the loss is computed per micro-batch split (micro sequence), and a document can span a split boundary. If each split returned its own document count as the denominator, documents straddling a boundary would be counted twice. That's why num_docs is computed only on micro_sequence_index == 0 and
  all_reduced — to count each document exactly once.

The current design handles this by precomputing num_docs at the batch level before splitting, and having the loss reference it by name via denominator_batch_field — essentially "use this precomputed scalar as my denominator" rather than computing it from within the split. Returning a denominator from the loss function directly would only work cleanly for
  quantities that are local to a split (like num_tokens or num_labels), but num_docs inherently crosses split boundaries so it can't be accumulated that way easily.

jlamypoirier: In either case what matters is `sum(numerator)/sum(denominator)`, so as long as the returned denominators sum to the right value (as in the valuers you're using) things still work

Anyway in #477 I'm precomputing `sum(denominator)` so we simply return `numerator/sum(denominator)` in losses and the reduction is a simple sum.

bigximik:
superseded by other PRs

PR #502, opened 2026-04-29
DeepSpeed parity hacks

bigximik:
## Summary

This PR adds GSPO loss to fast-LLM along with a suite of supporting fixes that together achieve full metric and training-trajectory parity with DeepSpeed's GRPO/GSPO implementation. Targets the `grpo-metrics` branch. Six logical units:

### 1. GSPO loss (sequence-level IS-ratio clipping)

Implements GSPO as an alternative policy-gradient loss alongside the existing per-token GRPO clipping. Controlled via `LanguageModelGRPOLossConfig.policy_loss = "gspo"`.

- New `fused_gspo_loss_forward_backward` kernel: computes per-segment geometric-mean log-ratio `R_s`, clips at `[1−ε_low, 1+ε_high]`, and applies `R_s × A_s` as a uniform per-token gradient within each segment. An `all_reduce(SUM)` over sequence-data-parallel ranks aggregates `(lrn_sum, adv_sum, tok_count)` before clipping so the ratio is correct under sequence parallelism.
- New `document_index` data field and `LanguageModelKwargs.document_index` kwarg constant to route per-token segment membership through the data pipeline.
- 8 unit tests in `tests/layers/test_gspo_loss.py` (single-segment, packed sequences, ratio=1 equivalence, clipping, masking, SDP mock, gradient finite-diff, independence from per-token metrics).

### 2. Dynamic `docs_per_step` accumulation

Replaces static `depth_first_micro_batches` with a runtime document-count target — matching DeepSpeed's `gradient_accumulation_passes` semantics for RL (where each microbatch holds one rollout).

- `ScheduleConfig.docs_per_step`: when >0, `Trainer._prefetch_to_doc_target` fetches microbatches one at a time, all-reduces the per-microbatch document count, and stops once the global total ≥ `docs_per_step`. The final step total is broadcast to all inputs so the normalisation denominator is consistent.
- `Trainer._get_or_build_schedule` builds and caches a per-N `Schedule` with `_depth_first_override = N // breadth_first_micro_batches`, so the existing schedule machinery is reused without changes to the runner.
- `Schedule._eff_{depth_first,sequential,num_inputs}` properties expose the effective values for a given override.
- 13 unit tests in `tests/layers/test_docs_per_step.py`.

### 3. `normalize_by_documents`

Adds a `normalize_by_documents` flag to `LanguageModelGRPOLossConfig`. When `True`, both the GRPO and GSPO paths divide the loss by `num_documents_in_batch` (the step-level rollout count) rather than the token count. Matches DeepSpeed's normalization where `tokens_weights = 1 / batch_size`.

### 4. Temperature scaling for IS ratio parity (moved to #516)

Adds a `temperature` field to `LanguageModelGRPOLossConfig`. When set to match the actor's sampling temperature (e.g. 0.7), new log-probabilities are computed at the same temperature as the stored old log-probabilities from vLLM, so the IS ratio starts near 1.0 at step 0 instead of ~1.08. Implementation: `_effective_logits_scale = logits_scale_factor / temperature`, substituted at all three call-sites in `_forward_backward`. Default `temperature=1.0` preserves existing behaviour exactly.

### 5. `fp32_lm_head` precision fix (matches vLLM's `bf16_last_layer_fp32`)

Adds a `fp32_lm_head` flag (default `False`) on `LanguageModelHeadConfig`. When `True`, the LM head's logits computation upcasts both input and weight to FP32 before the linear projection, matching vLLM's `bf16_last_layer_fp32` quantization. This ensures the trainer computes log-probabilities at the same numerical precision as the actor's sampling, so `new_logprobs ≈ old_logprobs` at step 0 (IS ratio at training start ≈ 1.0, not artificially inflated by precision mismatch).

- Commit `d8cb9ef5`: introduces the flag, upcasts input/weight, casts back to BF16 before downstream consumption.
- Commit `0f90f20b`: fixes the gradient flow when `fp32_lm_head=True`. The detached FP32 weight copy has `requires_grad=False`, which makes `output_parallel_linear_backward` skip writing to the original weight's `grad_buffer`. We restore the FSDP gradient contract by computing `grad_weight = grad.t() @ saved_input` explicitly and accumulating into the BF16 param's `grad_buffer` via `accumulate_gradient`.

### 6. Decoupled loss/gradient divisors and SDP loss double-counting fix

Even with `normalize_by_documents=true`, fast-LLM's reported `grad_norm` was ~1024× larger than DeepSpeed's, causing the default `gradient_norm_clipping=0.3` to over-clip by ~500× and making training ~10 reward points slower than DS GSPO at the same step count. Two issues, fixed in commit `557a3c4c`:

**Asymmetric loss/gradient scaling in DS**:
- DS loss reported uses `/batch_size` once (via `tokens_weights = 1/batch_size`, `pipelinerl/finetune/rl/__init__.py:246`).
- DS gradient buffer has an ADDITIONAL `/(gas × world_size)` factor from `scale_wrt_gas=True` in `engine.backward()` (`deepspeed/runtime/engine.py:1995-1996`) and `tensor.div_(world_sz)` in `reduce_scatter_coalesced` (`deepspeed/runtime/comm/coalesced_collectives.py:124`).
- For `samples_per_microbatch=1` (PipelineRL standard), `gas × world_size = batch_size`, so the gradient buffer effectively has `1/batch_size²` while the loss metric has `1/batch_size`.

Fast-LLM cancels DS's `/(gas × world_size)` factor structurally via `grad_output = data_parallel × grad_scale` (`runner.py:318`) interacting with FSDP's RS-AVG over `data_parallel` ranks (`fsdp.py:396`). So we need to apply the second `1/batch_size` factor explicitly only to the gradient — keeping the loss metric matched to DS.

Fix: add a `grad_divisor` parameter to `fused_gspo_loss_forward_backward`, `fused_grpo_loss_forward_backward`, and `triton_grpo_loss_forward_backward`. When `normalize_by_documents=true`:
- loss divisor = `num_documents_in_batch` (matches DS `rl/loss`)
- gradient divisor = `num_documents_in_batch²` (matches DS `grad_norm`)

Independent of TP/PP/SDP/DP parallelism and microbatching schedule, because `batch_size` is invariant under all of them.

**SDP loss double-counting**:
After the SDP allreduce of `lrn_sum/adv_sum/tok_sum` in `fused_gspo_loss_forward_backward`, both SDP ranks compute IDENTICAL per-segment loss values. When `LossDef.reduce` SUMs across `data_group` (which includes SDP ranks), the loss metric is double-counted by `sdp_size`. The gradient is NOT double-counted — each SDP rank contributes gradient from its own LOCAL tokens, with different contributions for different tokens of the same segment.

Fix: divide loss by `sdp_size` when `sdp_group` is active. Gradient unaffected.

## Verification

End-to-end 7B math run on 4 nodes, GSPO, `gradient_norm_clipping=0.3` (default), `normalize_by_documents=true`, `fp32_lm_head=true`, `temperature=0.7`:

| Metric | Before unit-6 fix | After unit-6 fix | DS GSPO reference |
|--------|-------------------|------------------|-------------------|
| step 1 `grad_norm` | 141 (1000× DS) | **0.135** | 0.145 |
| step 1 `lm_head_loss` | -13.7 | ~-1.7 magnitude | -1.7 |
| step 1 `clip_coeff` | 0.002 (severe over-clip) | **1.000** (no clip) | no clip |
| step 50 `newlp` | trapped at -0.17 | **-0.103** | -0.105 |

newlp trajectory tracks DS step-by-step. Both systems show same gradient-spike pattern during warmup ramp-up at steps 14-20 (DS step 16 grad_norm=6.365, fast-LLM step 15=9.005). Match within data variance.

## Test plan

- [x] `pytest tests/layers/test_gspo_loss.py` — GSPO unit tests pass
- [x] `pytest tests/layers/test_docs_per_step.py` — docs_per_step unit tests pass
- [x] `pytest tests/layers/test_lm_losses.py` — existing GRPO loss + per-token metrics tests unaffected (the metrics tests previously in `test_grpo_metrics.py` moved into this file on the base branch)
- [x] End-to-end: 4-node Qwen2.5-7B math run with full config (`docs_per_step=1024`, `temperature=0.7`, `normalize_by_documents=true`, `fp32_lm_head=true`, default `gradient_norm_clipping=0.3`) — grad_norm matches DS at step 1, training trajectory matches DS step-by-step through step 50+ (ongoing run validates through step ~410).


jlamypoirier:
# Coarse review — pass 1 of 2

Reviewed `git diff origin/grpo-metrics...origin/gspo` (head `15ae8d66`) against the base branch `grpo-metrics`. Structural / correctness pass only — naming, comments, and formatting nits go to `/review-fine`.

**1.** `LanguageModelGRPOLossConfig.policy_loss: str` at `fast_llm/layers/language_model/loss/config.py:208-212` violates the loss-config dispatch convention. Sibling losses (`label`, `distillation`, `dpo`, `z_loss`, `grpo`) are all registered via `@config_class(dynamic_type={LanguageModelLossConfig: "..."})` (`fast_llm/layers/language_model/loss/config.py:78-204`). Add a `@config_class(dynamic_type={LanguageModelLossConfig: "gspo"})` sibling — extracting a shared base for the fields/code GSPO and GRPO actually have in common — instead of a stringly-typed switch inside `LanguageModelGRPOLoss._forward_backward`.

**2.** The fp32_lm_head manual weight-grad path at `fast_llm/layers/language_model/head.py:288-298` reaches into `output_parallel_linear_forward`'s context tuple by positional index (`context[0]`, `context[3]`, `context[4]`) with only a comment as documentation. The forward returns 9 fields (`fast_llm/functional/linear.py:136-145`); a future reorder will silently miscompute the weight gradient or feed `gather_op` the wrong arg. Lift the FP32-weight handling into `output_parallel_linear_backward` (e.g. accept an explicit `weight` override or a "compute-grad-against-this-other-param" hook) so the head doesn't replicate linear-backward internals.

**3.** Same fp32_lm_head block re-gathers `saved_input` on the sequence-parallel path even though `output_parallel_linear_backward` already gathers it internally (`fast_llm/functional/linear.py:152-157`). On the FP32 path that's two gather collectives per backward instead of one. Fix together with item 2 — once the weight grad lives in `output_parallel_linear_backward`, the gathered `input1` is already available there.

**4.** `Schedule.__init__(_depth_first_override=...)` at `fast_llm/engine/schedule/schedule.py:118-121` is a private back-channel that shadows `config.depth_first_micro_batches` via a parallel `_eff_depth_first` / `_eff_sequential_micro_batches` / `_eff_num_inputs` API (`fast_llm/engine/schedule/schedule.py:160-174`) used in five places. Make `depth_first_micro_batches` a regular constructor argument (default = `config.depth_first_micro_batches`), drop the `_eff_*` shadow, and let the existing `self._config.X` reads resolve through the constructor-stored value.

**5.** GSPO and GRPO call sites in `LanguageModelGRPOLoss._forward_backward` at `fast_llm/layers/language_model/loss/grpo.py:101-133` duplicate the kernel-kwargs block almost verbatim (only the GSPO-specific `document_index` and `sdp_group` differ). Build one shared kwargs dict and dispatch the kernel choice on top, or — better — move the dispatch to the subclass introduced in item 1 so each kernel call lives in its own override.

**6.** `_prefetch_to_doc_target` at `fast_llm/engine/training/trainer.py:170-174` hard-fails via `Assert.eq(len(buffer) % bfmb, 0, ...)` whenever the natural document-target stopping point isn't `breadth_first_micro_batches`-aligned. With a streaming dataloader and any `bfmb > 1`, this crashes mid-step on data-dependent boundaries. Either keep fetching microbatches until alignment is reached (rounding `total_docs` upward), or validate the combination at config time so it can't happen at runtime.

**7.** `self._preprocessing_config = preprocessing_config` at `fast_llm/engine/training/trainer.py:118` stores a value nothing else in the diff (or in the file) reads. Drop the assignment.

**8.** `self._schedule = Schedule(...)` at `fast_llm/engine/training/trainer.py:121-127` is built unconditionally but used only on the `docs_per_step == 0` path (line 281 vs 290). Either skip the build when `docs_per_step > 0`, or unify by feeding the static path through `_get_or_build_schedule(sequential_micro_batches)` and removing the duplicate construction site.

**9.** The SDP-loss double-counting fix at `fast_llm/layers/language_model/loss/grpo.py:481-482` (`loss = loss / world_size(sdp_group)`) is the most subtle correctness change in the PR and is not exercised by any test. `tests/layers/test_gspo_loss.py::test_sdp_mock` calls the kernel with `sdp_group=None` throughout and reconstructs the SDP semantics manually — the actual `if sdp_group is not None: loss = loss / sdp_size` branch never runs in tests. Add a multi-rank `torchrun`-spawned test asserting `loss(SDP=k) == loss(SDP=1)` on the same global batch.

**10.** No test exercises `Trainer._get_or_build_schedule` (`fast_llm/engine/training/trainer.py:146-158`) — only its `_eff_*` byproduct properties on `Schedule`. The cache key behavior, the actual schedule built with the override, and reuse across iterations are untested. Add a unit test that asks for several `n_microbatches` values, asserts caching, and checks the resulting schedules' `_eff_*` values.

**11.** No test covers the fp32_lm_head gradient path at `fast_llm/layers/language_model/head.py:288-298`, especially the `cross_entropy_splits > 1` case where `accumulate_gradient(self.output_weights, ...)` is called once per split and relies on `param_grad_is_zero` being correctly toggled across calls. Add a unit test that runs forward+backward with `fp32_lm_head=True` (and again with `cross_entropy_splits=2`) and compares `output_weights.grad_buffer` to a non-FP32 reference within tolerance.

**12.** `int(document_index.max().item())` at `fast_llm/layers/language_model/loss/grpo.py:440` triggers a host sync on every GSPO microbatch, and the SDP `n_segs` `all_reduce(MAX)` at line 443 is a separate collective from the `lrn_sum`/`adv_sum`/`tok_sum` `all_reduce(SUM)` at lines 461-463. Compute `n_segs` once at batch construction (it's already known from `length_cumsum` in `_get_label_counts`), thread it through `kwargs`, and drop both syncs.

**13.** `torch.ones(masked_doc_ids.numel(), device=logits.device)` at `fast_llm/layers/language_model/loss/grpo.py:457` allocates a fresh ones-tensor every forward, only to feed `index_add_`. Use `torch.bincount(masked_doc_ids, minlength=n_segs).to(log_ratio.dtype)` — no allocation, no separate `index_add` pass.

**14.** `LanguageModelGRPOLossConfig.normalize_by_documents`'s description at `fast_llm/layers/language_model/loss/config.py:230-236` says "Set to True when using docs_per_step for full DS parity" but nothing in `_validate` enforces or even cross-references the two flags (which live in different configs). Either add a cross-config consistency check on the trainer/experiment config, or drop the load-bearing-sounding sentence from the field doc.

## Notes

- The decision-rich comment block at `fast_llm/layers/language_model/loss/grpo.py:78-91` (the loss-vs-grad divisor derivation) belongs in a design doc or commit-message reference rather than in the hot-path forward — but it's the kind of "why" `CLAUDE.md` permits, and removing it would lose hard-won context. Flagging only as a maintenance hazard: the line numbers it cites in DeepSpeed will drift.

- `LanguageModelGRPOLoss.get_preprocessing_config` at `fast_llm/layers/language_model/loss/grpo.py:240-243` adds `return_document_index` only when `policy_loss == "gspo"`. If item 1 is acted on (sibling subclass), the override moves naturally into the GSPO subclass and the conditional disappears.

- The `_StubTrainer` pattern in `tests/layers/test_docs_per_step.py:996-1001` (borrowing `_prefetch_to_doc_target` directly off `Trainer`) silently breaks if the method gains an attribute access not stubbed on the fake. Prefer extracting the prefetch logic into a free function that takes its dependencies as arguments — the test then exercises the same function the trainer calls.

jlamypoirier:
About the features
1. Looks useful.
2. As per out previous discussion, likely unnecessary.
3. My understanding is that this is the fundamental point of GSPO
4. Sounds good, but need to rename to logits_scale_factor for consistency.
5. Looks unnecessary. Effect on logprob ratio should be negligible
6. Deepspeed is clearly the wrong one here. Instead of reproducing the bug, we should use a more appropriate value for gradient clipping.

Do we know what was actually responsible for the inconsistency with Deepspeed? In my opinion 6 is the most suspicious. 1/3 is equally problematic but looks more like a config / scoping issue (were we always using GSPO on the deepspeed side? Did we ever compare using GRPO?). 4 look real. but minor, 2 and 5 look negligible.

jlamypoirier:
Feature 4 (temperature for IS ratio) was renamed to a per-loss `logits_scale_factor` field on the base loss config and merged in #516. The `temperature` field on this branch has been removed in favor of it.

jlamypoirier:
#520

PR #526, opened 2026-05-27
Add fp32_lm_head flag for vLLM precision parity

jlamypoirier:
## Summary

Adds an `fp32_lm_head` field on `LanguageModelHeadConfig`. When `True`, the LM head linear's input and weight are upcast to FP32 before the matmul, matching vLLM's `bf16_last_layer_fp32` quantization. This lets the trainer compute log-probabilities at the same numerical precision as the actor's sampling, so the importance-sampling ratio starts near 1.0 instead of being artificially inflated by a trainer/actor precision mismatch.

The detached FP32 weight has `requires_grad=False`, which makes `output_parallel_linear_backward` skip the weight-grad path. The FSDP gradient contract is restored by computing `grad_weight = grad.t() @ saved_input` explicitly and accumulating into the original BF16 param's `grad_buffer` via `accumulate_gradient`.

Off by default — disabled path is byte-identical to before.

## Test plan

- [x] `pytest tests/layers/test_lm_head.py` — passes

---

Originally part of #502.


jlamypoirier:
Claude Opus 4.8 (for @joel-lamy-poirier): Folding the `fp32_lm_head` changes into #520, which is now retargeted to `main` and kept as the single PR to main. `jlp_fp32_lm_head` is an ancestor of `jlp_rl_features`, so all of these changes are included in #520's diff. Closing in favor of #520.

PR #473, opened 2026-03-07
Allow no bos for Qwen

shruthan:
# ✨ Description
Allows data prep with tokenizers that don't have a bos token (like the Qwen family). 

Select all that apply:

- [x] 🐛 **Bug fix** (non-breaking change that addresses a specific issue)
- [ ] 🚀 **New feature** (non-breaking change that adds functionality)
- [ ] ⚠️ **Breaking change** (a change that could affect existing functionality)
- [ ] 📈 **Performance improvement/optimization** (improves speed, memory usage, or efficiency)
- [ ] 🛠️ **Code refactor** (non-functional changes that improve code readability, structure, etc.)
- [ ] 📦 **Dependency bump** (updates dependencies, including Dockerfile or package changes)
- [ ] 📝 **Documentation change** (updates documentation, including new content or typo fixes)
- [ ] 🔧 **Infrastructure/Build change** (affects build process, CI/CD, or dependencies)

## 📝 Changes

List the key changes introduced in this PR:

1. Change A
2. Change B

## ✅ Checklist

Make sure the following tasks are completed before submitting the PR:

### General

- [ ] 📜 I have read and followed the [contributing guidelines](https://servicenow.github.io/Fast-LLM/contributing/contributing).
- [ ] 🏷️ I am using a clear and descriptive PR title that summarizes the key change or feature introduced.
- [ ] 🎉 The functionality is complete, and I have tested the changes.
- [ ] 📝 I have updated the documentation if needed.
- [ ] ⚠️ The change does not introduce any new issues (e.g., runtime warnings, type checker errors, linting problems, unhandled edge cases).
- [ ] 🧩 I have commented my code, especially in hard-to-understand areas.

### Dependencies and Configuration

- [ ] 🐋 I have updated the Docker configuration or dependencies, if applicable.
- [ ] 🔄 I have ensured compatibility with the existing setup after dependency changes.

### Testing

- [ ] 🧪 I have added or updated tests to cover my changes.
- [ ] ✔️ New and existing tests pass locally with my changes.
- [ ] 🚦 I have tested these changes on GPUs and verified training stability.
- [ ] 🏋️ I have tested the changes on realistic training workloads, if applicable.

### Performance Impact

- [ ] 📊 I have run benchmarks where applicable to evaluate the performance impact.
- [ ] ✅ The benchmarks show no performance regression.
- [ ] 🚀 The benchmarks indicate a potential performance improvement.
- [ ] ⚠️ The benchmarks indicate a potential performance degradation.
[...]

PR #535, opened 2026-06-05
Add configurable train_on_eos for conversation data preparation

jlamypoirier:
_Authored by Claude Opus 4.8 (with @jlamypoirier)._

Splits out — as a standalone, opt-in flag — the loss-masking change that was bundled into #473.

## What

Add `train_on_eos: bool = False` to `ConversationSourceConfig`. It controls whether the end-of-sequence token that `tokenize_chat` appends after the final message is included in the training loss:

- `false` (default): the appended EOS is masked from the loss — **unchanged behavior**.
- `true`: the appended EOS becomes a training target.

Threaded through `Tokenizer.tokenize_chat` as a `train_on_eos` parameter. Per-turn terminators emitted by the chat template (e.g. ChatML `<|im_end|>`) are unaffected — they remain governed by the template's `{% generation %}` markers. This flag only touches the single sequence-terminating EOS that the tokenizer appends when no EOS already appears in the conversation.

## Why

Masking the terminal EOS means the model never gets a loss signal to emit end-of-sequence — a well-known cause of models that don't stop generating at inference. Training on the final/assistant EOS is the common recommendation, and frameworks expose it as a knob (Axolotl `train_on_eos: turn|all|last`; TRL `assistant_only_loss`, which includes the assistant turn's EOS). It's also an asymmetry within Fast-LLM today: the document path already trains on its appended EOS (unmasked tokens all contribute to the loss), while the conversation path masks it. This flag lets conversation prep opt into the same behavior.

Kept off by default so existing datasets' loss masking is unchanged.

## Testing

`tests/data/test_tokenizer.py` (incl. the new `test_tokenize_chat_train_on_eos`) and `test_preparator.py` pass (41). The new test asserts that `train_on_eos` changes only the appended EOS's loss mask, not the tokens.

## Note

Touches the same `tokenize_chat` call site as #534 (no-BOS prep); whichever merges first, the other needs a one-line rebase.

### DIFF
--- a/fast_llm/data/preparation/gpt_memmap/config.py
+++ b/fast_llm/data/preparation/gpt_memmap/config.py
@@ -170,6 +170,12 @@ class ConversationSourceConfig(LanguageModelSourceConfig):
         desc="Field containing the conversation messages list. Each message should have 'role' and 'content' keys.",
         hint=FieldHint.core,
     )
+    train_on_eos: bool = Field(
+        default=False,
+        desc="Include the end-of-sequence token appended after the final message in the training loss."
+        " When disabled, that token is masked from the loss.",
+        hint=FieldHint.optional,
+    )
 
     @functools.cached_property
     def columns(self) -> list[str]:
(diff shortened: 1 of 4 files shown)


jlamypoirier:
Claude Opus 4.8 note: Folded into #534 along with the no-BOS data-prep changes (one PR instead of two, per maintainer preference). The `train_on_eos` flag is unchanged. Closing in favor of #534.

PR #553, opened 2026-07-07
RL diagnostics: documents-seen clock, model-version broadcast, reward + per-token model_version, data-wait

jlamypoirier:
_Claude Opus 4.8, on behalf of @jlamypoirier._

## Summary

Consolidates the remaining RL diagnostics on the Fast-LLM trainer side so the trainer is a single,
reliable source of truth for GSPO/GRPO runs — logged from the exact tensors each step trained on.

The config unification + GSPO/GRPO policy metrics landed separately in #555; **this PR is rebased on
top of it** and adds the document clock, the version broadcast, the reward / per-token-version data, and
the data-wait timer. Paired PipelineRL changes (raw-reward forwarding + document-count model version)
live in a separate PR against PipelineRL's `fast-llm` branch. Backward-compatible with the current producer.

## Changes

- **`documents_seen` / `num_documents`.** Reimplemented on `main`, independent of `docs_per_step`, from
  the per-step DP-summed document count (the loss-normalization divisor). Persisted in checkpoint
  metadata. Offered as an alternative W&B x-axis via a lazily-declared `define_metric` (RL runs only;
  the global step is unchanged).
- **Document-count model version.** `weights_ready` now carries `document_count` (= `documents_seen`)
  alongside the existing `step`, so consumers can measure staleness in documents (DeepSpeed-aligned).
  Keeping `step` makes this backward-compatible.
- **`reward` and per-token `model_version`.** Two optional per-token fields threaded through the RL
  streaming schema and data pipeline (reusing `TokenDataDocument`/`TokenDataBatch`); both optional and
  guarded on presence. `reward` is the raw reward (distinct from the group-relative `advantage`). The
  shared policy-gradient loss logs mean/max/min of each when present (folded into the `PolicyMetricsLevel`
  metric set from #555). Staleness is `documents_seen − model_version`, derivable from the logged stats.
- **`data_wait_time_ms`.** The schedule runner already timed data-loader waits but discarded the value;
  it is now logged as an input-starvation metric.

## Tests

- The reward/version registration is checked against an independent plain-Python reference
  (`test_policy_data_metrics`), and the streaming schema round-trips the new fields (`test_streaming_dataset`).
- Rebased onto #555 (now in `main`); CI validates the full suite.

### DIFF
--- a/fast_llm/data/dataset/streaming.py
+++ b/fast_llm/data/dataset/streaming.py
@@ -31,6 +31,10 @@ class RedisStreamingDocumentData(Config):
     rejected_span: tuple[int, int] | None = Field(default=None)
     advantage: float | None = Field(default=None)
     old_log_probabilities: torch.Tensor | None = Field(default=None)
+    # Raw (un-normalized) reward, a per-rollout scalar (broadcast per-token like `advantage`).
+    reward: float | None = Field(default=None)
+    # Model version each token was generated under (documents-seen units), one per token.
+    model_version: torch.Tensor | None = Field(default=None)
 
     def _validate(self):
         # Decode message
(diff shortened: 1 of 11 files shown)


jlamypoirier:
Claude Opus 4.8 note (on behalf of @jlamypoirier): closing — this umbrella has been fully broken down into independent PRs:

- #555 — Unify policy-loss metrics config and add GSPO metrics
- #559 — Log documents_seen / num_documents and offer it as a W&B x-axis
- #560 — Log weight-sync time as a streaming-trainer metric
- #561 — Broadcast documents_seen as the model version in weights_ready
- #562 — Carry reward and per-token model_version, log reward/version metrics
- #564 — Log data-loader wait time as an input-starvation metric

#555/#559/#560/#561/#562 are merged; #564 (the final piece) is open. Nothing remains here.

